PR #1, opened 2020-11-11
Make HDF5 to Zarr translation self-sufficient

ajelenak:
This version does not depend on a custom zarr package compared to https://gist.github.com/ajelenak/80354a95b449cedea5cca508004f97a9.


martindurant:
Thanks. I have merged and cleaned things up a bit - sorry for messing up the git history.

PR #3, opened 2020-11-11
Define spec and make HDF conform

martindurant:


martindurant:
Needs https://github.com/intake/filesystem_spec/pull/470

Issue #4, opened 2020-11-18
two offsets or offset + size?


rabernat:
In most of original zarr issue discussion (https://github.com/zarr-developers/zarr-python/issues/556), offsets were specified with `offset, size`. In this spec, you are doing `start, stop`. It seems like `offset, size` will probably be slightly smaller as a text representation. I'm curious how you made that choice?

martindurant:
I believe it was the API of the "fetch" functions littered throughout fsspec. So... mental convenience.

manics:
I think `offset`, `size` is also less likely to cause confusion across different languages and systems. For example Python uses `inclusive-exclusive` ranges while [HTTP partial requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#Requesting_a_specific_range_from_a_server) are `inclusive-inclusive`.

PR #5, opened 2020-11-18
Change to offset, size

rabernat:
As discussed in #4, the preference from the community seems to be for `offset, size` instead of `start, stop` for encoding item locations within a file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,12 @@ Spec for the structure required by ReferenceFileSystem:
 {
   "key0": "data",
   "key1": {
-    ["protocol://target_url", 100, 200]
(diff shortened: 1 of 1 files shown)


martindurant:
Hoping to get many suggestions like this one, to arrive as a sensible concrete spec!

It would be handy to update the HDF crawler implementation as the spec changes (and later ReferenceFileSystem in fsspec) so that things don't get out of sync too much.

rsignell-usgs:
@rabernat, Smaller representation and easier to socialize?   Sounds good to me!

martindurant:
OK merging. Spec before implementation :)

PR #12, opened 2020-11-29
Adding example notebook with binder link

rsignell-usgs:
Here is the [rendered notebook](https://nbviewer.jupyter.org/gist/rsignell-usgs/cd7821fa728512853e75feb1912058e0) example.
 
I've been saying that reading HDF5  is not significantly different than reading zarr, but it's actually about 20% slower for this example.   With our original test, the times were not significantly different.   @martindurant, Would you expect our new mapping approach to be slower?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,5 +29,5 @@ For example, Zarr data in this proposed spec might be represented as:
 ```
 
 
-Run a notebook example comparing reading HDF5 using this approach vs. native Zarr format:
-[![Binder](https://aws-uswest2-binder.pangeo.io/badge_logo.svg)](https://aws-uswest2-binder.pangeo.io/v2/gh/rsignell-usgs/fsspec-reference-maker/example?filepath=examples%2Fike_intake.ipynb)
+
+[![Binder](https://aws-uswest2-binder.pangeo.io/badge_logo.svg)](https://aws-uswest2-binder.pangeo.io/v2/gh/rsignell-usgs/fsspec-reference-maker/example?filepath=examples%2Fike_intake.ipynb): Run a notebook example comparing reading HDF5 using this approach vs. native Zarr format
--- a/README.md
+++ b/README.md
@@ -29,5 +29,4 @@ For example, Zarr data in this proposed spec might be represented as:
 ```
 
 
-Click here to run a notebook comparing reading HDF5 with Zarr, using this approach:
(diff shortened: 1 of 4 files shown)


rsignell-usgs:
@martindurant , any thoughts?  
@kwilcox, give this a try!

martindurant:
I'll merge this, so that the binder link appears and it is easier to play with.

martindurant:
Yep, I think it's a nice demonstration. 
I ran it with no problems now: 39s for HDF and 30s for zarr. It may be that the second one is faster just because it is second (workers need to import pandas, etc). When I ran it again, HDF came in at 28s.

Is it worth showing the contents of the catalogue or mentioning Intake here? I suppose the audience already knows, but it's nice to show that this works either way transparently.

martindurant:
(note that it comes up as "classic" notebook for me, which is totally fine, but you could have a jlab interface with the dask panels populated for a bit more splash)

kwilcox:
> Is it worth showing the contents of the catalogue or mentioning Intake here? I suppose the audience already knows, but it's nice to show that this works either way transparently.

I had to spend some time digging into `intake_xarray.xzarr.ZarrSource` and the parameters being passed as part of the catalog config to figure out what was actually happening, I wouldn't mind the explanation!

rsignell-usgs:
Here are my latest results, running the 2 cases twice to make sure the cluster was spun up.  So @martindurant I guess you might be right.  These are my fastest times yet:

martindurant:
> These are my fastest times yet!

Except you can get even better by using partitions containing multiple zarr chunks!

rsignell-usgs:
And I can do that in the catalog, right?   Can you provide (refresh) my memory?

martindurant:
Yes, you add a `chunks` key to the zarr args, like `{coord: size, ...}`, but I'm not sure what the right set of chunks would be.

martindurant:
I think you deleted your branch, and the binder link pointed to it, not to the main repo. 

My AWS credentials have disappeared, so I can't access it from here, but it would be something like 
```yaml
  ike-hdf5:
    driver: intake_xarray.xzarr.ZarrSource
    description: 'Hurricane Ike simulation in NetCDF4/HDF5 format'
    args:
      storage_options:
        ref_storage_args:
          requester_pays: true
        references: "s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d_offsets.json"
        target_options:
          requester_pays: true
        target_protocol: s3
      urlpath: "reference://"
      chunks:
        time: 30
```
where the number 30 is some multiple of the default chunk size that you get.

PR #16, opened 2020-12-29
Raise error for compact HDF5 datasets

ajelenak:
I added a RuntimeError exception when a compact HDF5 dataset is encountered during HDF5-to-Zarr translation until ReferenceFileSystem supports it. See this [comment](https://github.com/intake/fsspec-reference-maker/issues/14#issuecomment-747144520).

### DIFF
--- a/fsspec_reference_maker/hdf.py
+++ b/fsspec_reference_maker/hdf.py
@@ -98,8 +98,8 @@ class Hdf5ToZarr:
                 source = value.pop("source")["uri"]
                 for k, v in value.items():
                     ref[k] = (source, v["offset"], v["size"])
-            else:
-                ref[key] = value.decode()
+            # else:
+            #     ref[key] = value.decode()
         return ref
 
     def transfer_attrs(self, h5obj: Union[h5py.Dataset, h5py.Group],
@@ -176,9 +176,8 @@ class Hdf5ToZarr:
             if self._xr:
                 # Do this for xarray...
                 adims = self._get_array_dims(h5obj)
-                if adims:
-                    za.attrs['_ARRAY_DIMENSIONS'] = adims
-                    lggr.debug(f'_ARRAY_DIMENSIONS = {adims}')
+                za.attrs['_ARRAY_DIMENSIONS'] = adims
+                lggr.debug(f'_ARRAY_DIMENSIONS = {adims}')
 
             # Store chunk location metadata...
             if cinfo:
--- a/fsspec_reference_maker/hdf.py
+++ b/fsspec_reference_maker/hdf.py
@@ -1,11 +1,10 @@
+from typing import Union, BinaryIO
 import logging
 from urllib.parse import urlparse, urlunparse
 import numpy as np
 import h5py
 import zarr
-import json
 from zarr.meta import encode_fill_value
-from zarr.storage import _path_to_prefix
 from numcodecs import Zlib
 import fsspec
 from zarr.util import json_dumps
@@ -15,6 +14,7 @@ chunks_meta_key = '.zchunkstore'
 lggr = logging.getLogger('h5-to-zarr')
 chunks_meta_key = ".zchunkstore"
 
+
 def _path_to_prefix(path):
     # assume path already normalized
     if path:
@@ -22,6 +22,7 @@ def _path_to_prefix(path):
     else:
         prefix = ''
     return prefix
+
 
 def chunks_info(zarray, chunks_loc):
     """Store chunks location information for a Zarr array.
@@ -39,7 +40,6 @@ def chunks_info(zarray, chunks_loc):
         raise ValueError(
             f'{chunks_loc["source"]}: Chunk source information incomplete')
 
-
     key = _path_to_prefix(zarray.path) + chunks_meta_key
     chunks_meta = dict()
     for k, v in chunks_loc.items():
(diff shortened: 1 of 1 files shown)


fsspec_reference_maker/hdf.py
```diff
@@ -113,12 +115,14 @@ def transfer_attrs(self, h5obj, zobj):
...
             if n in ('REFERENCE_LIST', 'DIMENSION_LIST'):
-                    continue
+                continue
 
             # Fix some attribute values to avoid JSON encoding exceptions...
             if isinstance(v, bytes):
                 v = v.decode('utf-8')
             elif isinstance(v, (np.ndarray, np.number)):
+                if v.dtype.kind == 'S':
+                    v = v.astype(str)
```
martindurant, line 125: Same as `np.char.decode()`?

fsspec_reference_maker/hdf.py
```diff
@@ -131,12 +135,19 @@ def transfer_attrs(self, h5obj, zobj):
                 zobj.attrs[n] = v
             except TypeError:
                 print(f'Caught TypeError: {n}@{h5obj.name} = {v} ({type(v)})')
+                lggr.exception(' ')
```
martindurant, line 138: I agree to prefer logging over the print on the line above, which should be removed/combined with this

fsspec_reference_maker/hdf.py
```diff
@@ -164,8 +175,9 @@ def translator(self, name, h5obj):
             if self._xr:
                 # Do this for xarray...
                 adims = self._get_array_dims(h5obj)
-                za.attrs['_ARRAY_DIMENSIONS'] = adims
-                lggr.debug(f'_ARRAY_DIMENSIONS = {adims}')
+                if adims:
```
martindurant, line 178: else?

fsspec_reference_maker/hdf.py
```diff
@@ -257,40 +269,10 @@ def storage_info(self, dset):
                 blob = dsid.get_chunk_info(index)
                 key = tuple(
                     [a // b for a, b in zip(blob.chunk_offset, chunk_size)])
-                stinfo[key] = {'offset': blob.byte_offset,
-                               'size': blob.size}
+                stinfo[key] = {'offset': blob.byte_offset, 'size': blob.size}
             return stinfo
 
 
-def chunks_info(zarray, chunks_loc):
```
martindurant, line 265: This is called in `tanslator()` under `if cinfo:`

fsspec_reference_maker/hdf.py
```diff
@@ -71,14 +72,14 @@ class Hdf5ToZarr:
...
-    def __init__(self, h5f, xarray=False):
+    def __init__(self, h5f: Union[str, BinaryIO], xarray: bool = False):
         # Open HDF5 file in read mode...
         lggr.debug(f'HDF5 file: {h5f}')
         lggr.debug(f'xarray: {xarray}')
         self._h5f = h5py.File(h5f, mode='r')
         self._xr = xarray
 
-        self.store = {}
+        self.store = zarr.MemoryStore()
```
martindurant, line 82: Why change this?

martindurant (commented):
Note that this PR does more than stated above, with some refactoring and the appearance of typing.

fsspec_reference_maker/hdf.py
ajelenak: The print statement was leftover from debugging. Removed.

fsspec_reference_maker/hdf.py
ajelenak: Is there a need for empty _ARRAY_DIMENSIONS_ attribute? My understanding is that no need for _ARRAY_DIMENSIONS_ if no HDF5 dimension scales.

fsspec_reference_maker/hdf.py
ajelenak: The `chunks_info()` function was duplicated in the code so I removed one copy.

fsspec_reference_maker/hdf.py
ajelenak: Looked nicer, more explicit to me. Returned to the ordinary dict.

fsspec_reference_maker/hdf.py
martindurant: I have no idea - but it seems like something we ought to eventually cover in a test.

My question was exactly: do we need an else block here? It would, I suppose, be for a scalar value.

ajelenak: I reverted back to always having this Zarr attribute. The description in the [xarray documentation](http://xarray.pydata.org/en/latest/internals.html#zarr-encoding-specification) seems to imply it is required.

fsspec_reference_maker/hdf.py
```diff
@@ -93,13 +135,20 @@ def transfer_attrs(self, h5obj, zobj):
...
-    def translator(self, name, h5obj):
+    def translator(self, name: str, h5obj: Union[h5py.Dataset, h5py.Group]):
         """Produce Zarr metadata for all groups and datasets in the HDF5 file.
         """
         if isinstance(h5obj, h5py.Dataset):
-            lggr.debug(f'Dataset: {h5obj.name}')
+            lggr.debug(f'HDF5 dataset: {h5obj.name}')
+            if h5obj.id.get_create_plist().get_layout() == h5py.h5d.COMPACT:
+                RuntimeError(
+                    f'Compact HDF5 datasets not yet supported: <{h5obj.name} '
```
martindurant, line 148: Please create an issue referencing this exception and your comment about this in the other issue. Hopefully we can cope with this soon.

Issue #13, opened 2020-12-14
Avoiding metadata bloat caused by many long URLs


rsignell-usgs:
I woke up wondering whether it would be possible to allow a variable to be defined in the referencefile spec. I'm worried about long s3 (or other) urls bloating the metadata. 

If we could do something like:
```
prefix='superlongurl/that/keeps/going/on/and/on/for/ever/to/some/dir'
"key1": {
    ["s3://$prefix/data001.nc", 10000, 100]
  }
```
we could make the bloat a lot smaller

martindurant:
Note that the current implementation in fsspec allows for a fallback default URL, for any entry where a URL is not specified, i.e., "[null, start, size]" (or `None` in python). That is not in the spec as reflected in this repo.

rsignell-usgs:
@martindurant , I did *not* know that.    How does one specify the fallback default URL?

martindurant:
argument `target=` to ReferenceFileSystem (would be under storage_options as part of an Intake data source).

rsignell-usgs:
In our [example metadata file ](s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d_offsets.json), replacing the explicit URLs with "$U" reduces the size by 50%.
```
"zeta/0.0": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 441750960, 1951430], "zeta/0.1": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 443702390, 2912202], "zeta/0.2": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 446614592, 2512331], "zeta/0.3": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 449126923, 3092137], "zeta/0.4": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 452219060, 5634126], "zeta/0.5": ["s3://pangeo-data-uswest2/esip/adcirc/adcirc_01d.nc", 457853186, 6748304],
```

martindurant:
Indeed, we might *not* use JSON as the only storage format in the long run. A binary format in which we can store the columns separately would give very nice compression for the case of repeated values (in the case of parquet, there is a specific encoding for this, although it is rarely used and poorly supported). Note that the key names also have a lot of repetition.

[...]

Issue #7, opened 2020-11-19
Templating and deterministic maps


rabernat:
In yesterday's meeting, we discussed the widespread desire for this spec to handle deterministic, evenly spaced-out chunks within a file, rather than explicitly enumerating each chunk and its offsets. We would need some sort of support for templates and symbolic expressions. For a 1D file, it could be something like

```
"{i}", ["s3://bucket/path/file.nc", "3200 * {i}", 3200]
```
or
```
"{i}", ["s3://bucket/path/file.nc", "{itemsize} * {i}", 3200]
```
where `{itemsize}` would get filled with 3200

For ND it would be harder because you need to know the array shape:
```
"{j}.{i}", ["s3://bucket/path/file.nc", "{itemsize} * ({j} * Nx + {i})", 3200]
```
I can't think of a clever way around that. 

Ideas? @manzt? @joshmoore?

martindurant:
In sketching this out, we may also want to consider:
- key names that are regex, which is a well-defined thing
- expressions are jinja-like, including support for functions
- expressions are simple, like "itemsize * i" (actually, if we know itemsize, I would replace it before saving)

I wonder, is there an overlap with the discussion about coordinates specified programatically in xarray?

To what extent does this the putative option of binary storage for offsets (i.e., not json, perhaps zarr itself) mitigate the problem, due to efficient compression of integers?

Sorry if that's too many thoughts for the limits of this issue.

rabernat:
Given the complexity, we might also consider punting on this and releasing an initial spec that requires explicit keys.

martindurant:
A spec can be added to easier than removed from ...

joshmoore:
Martin suggested my adding feedback I received from the @bioformats team here. I asked in a general way if there were anything needed for the spec to maximize the number of formats that we could support. There's hope that _some_ files for _some_ of the proprietary formats will be readable, but just taking the most prevalent (TIFF), there are a large number of edge cases that would require a callback function of some form. So a fourth argument:
```
(name, templated-size, templated-offset, processing-function)
```
since data that is interleaved, reversed, sample packed, etc. might need reversing, striding, etc. Details can be found in [TiffParser.getSamples](https://github.com/ome/bioformats/blob/develop/components/formats-bsd/src/loci/formats/tiff/TiffParser.java#L872). Obviously, this is in no way a MUST (all for KISS) but it would be interesting to hear if a minimal library of such functions would help in other domains as well.

[...]

Issue #8, opened 2020-11-19
versioning the spec


martindurant:
Once we have a complete spec of some sort, we should give it a version, which also is saved into the prescriptions, so that we can do backward-incompatible changes later

PR #17, opened 2021-03-01
Spec version 1

martindurant:


martindurant:
Fixes #7 
Fises #8
Fixes #13 

Ping @rabernat @rsignell-usgs @manzt @joshmoore

rabernat (commented):
Thanks so much for working on this @martindurant! I'm so excited about the possibilities.

My major concerns are the following:
- This is not really a "spec"; it's more of an example. A true spec would be more explicit about MUST, MAY, etc. and try to enumerate the full space of allowed possibilities .
- It seems like we are baking python language conventions into the spec. Is that the right choice?

martindurant:
> This is not really a "spec"; it's more of an example

I thought more like an example with explanations that can together be turned into a formal spec :) It could be encoded in a [jsonschema](https://json-schema.org/ ) and used for validation.

>  if you showed the complete representation that would be generated by the template

Good idea. Indeed, when implemented in code, this ought to be a test case.

> Do we want to link explicitly to a spec document URL

I think version should be a number, but it would be reasonable to also add a link/DOI, etc., in a different field. In the python implementation, we would probably not check this field.

> Is there a language-agnostic version of format strings we could point to?

I am happy to have one pointed out, but I believe python's is moderately simple and similar to other languages. Other popular template engines like jinja are also typically tied to a particular implementation.

> Putting python code in the spec seems problematic

The alternative is to come up with our own language... For the specific case of `range`, you could replace with a list literal, but would not want to do that for long lists.

martindurant:
This is the python version of the ["moustache"](http://mustache.github.io/) templating framework, which appears to have libs in many many languages: https://github.com/noahmorrison/chevron

README.md
```diff
@@ -28,7 +29,60 @@ For example, Zarr data in this proposed spec might be represented as:
...
+  templated parameters.
+    - key, url: generated key names and target URLs
+    - offset, length: to define the bytes range, will be converted to int
+    - additional named parameters: for each iterable found (i.e., returns successfully from `iter()`), creates a 
+      dimension of generated keys
+- refs: keys with either data or [url, offset, length]. The URL will be treated as a template if it contains 
+  "{" and "}".
+
+In the example, "key2" becomes ["protocol://long_text_template", ..] and "key3" becomes ["protocol://text", ..].
+Also contained will be keys "gen_ref0": ["protocol://long_text_template_0", 1000, 1000] to "gen_ref8":
```
ajelenak, line 81: Should it be `gen_key0` and `gen_key8` instead of `gen_ref...`?

martindurant: ... which is why it's good to have your spec be in code and tested too, right?

joshmoore:
Only comment on top of those given so far, I think, is that `data` doesn't seem to be specified. My _assumption_ is that it's the raw return value with no need for further lookup.

Regarding the Python-ism points, I could certainly see having a second implementation to keep us honest. I'd probably default to doing so in Java, but I won't start on that until I hear whether or not @manzt already has one in JS (which is what usually happens...)

martindurant:
> data doesn't seem to be specified. My assumption is that it's the raw return value with no need for further lookup.

Indeed. And that the remote chunks referenced are byte blocks too. To be included in JSON, these literals would have to be ascii (which is fine and normal for zarr). We could optionally introduce an encoding like b64 at this point.

As an interesting contrast, for something like the grib2 case, where we need to make a local file to pass to C, one could imagine using this kind of structure and specify a "literals processor" and/or "reference processor" somehow. Very useful, but I think beyond Version 1.

rabernat:
Perhaps we want to restrict what sort of python expressions can be used. Do we envision needing anything besides `range`? If not, we could do something like

```
"i": {"start": 0, "stop": 9}
```

martindurant:
> Do we envision needing anything besides range?

open to suggestions. By using a specific dict as you suggest, though, would hinder future expansion of possibilities, so I would still prefer `range` (with one, two or three arguments).  I originally had imagined allowing anything in python `builtins`, but indeed that would be hard to make language agnostic.

rabernat:
Then what about

```
"i": {"range": {"start": 0, "stop": 9}}
```

My point is that functions and arguments can be encoded as json and then translated to python, rather than putting python code as strings into the json document.

martindurant:
That syntax is OK with me.

manzt:
> This is the python version of the "moustache" templating framework, which appears to have libs in many many languages.

I think moustache is a good choice for this, echoing @rabernat concern with python-isms.

> I originally had imagined allowing anything in python builtins, but indeed that would be hard to make language agnostic. 

Ideally we could define the "set" of builtins that the spec may include in JSON like @rabernat's example. This would help clarify which builtins are absolutely necessary and as @joshmoore said "keep us honest" between implementations.

> Only comment on top of those given so far, I think, is that `data` doesn't seem to be specified. My assumption is that it's the raw return value with no need for further lookup.

Also somewhat unclear to me. Can `data` be both an ascii-encoded string that needs to be parsed as JSON or just JSON?

> I won't start on that until I hear whether or not @manzt already has one in JS (which is what usually happens...)

I've written a few ad-hoc pieces of code to parse the spec, but it's coupled with a custom zarr.js store. I can try to separate out pieces and work on something more "official" and reusable for the JS folks.

martindurant:
> I've written a few ad-hoc pieces of code to parse the spec, but it's coupled with a custom zarr.js store.

Since you don't have an fsspec to handle the references, I imagine a custom store is the only way to go. That's how this "pieces of HDF" was initially implemented too; but I am keen that the idea be transferrable beyond zarr if possible, at least in python.

manzt:
> Since you don't have an fsspec to handle the references, I imagine a custom store is the only way to go. 

Makes sense. I could image an `fsspec-reader-js` that is just a parser for spec and returns either `data` or a `ref` tuple for a given key, delegating how/if a data request should be made to someone else. I think for now (since I'll probably be one of few users), it makes sense to keep these together.

manzt:
As a side note, I'm not super familiar with `mustache`, but I'm not sure (don't think) expression evaluation is supported. So handing:
```
  "offset": "{(i + 1) * 1000}"
```
could be challenging.

ajelenak:
Checking whether this in the spec:
```json
{
    "refs": {
         "key0": "data",
    }
}
```
covers the use case below ([here](https://github.com/intake/fsspec-reference-maker/issues/14#issuecomment-747144520) for context):
```json
{
    "key0": "data:application/octet-stream;base64:<array data>"
}
```

martindurant:
Good point @ajelenak : this is not described. If the only two formats we consider are text (ascii) and b64-binary, then maybe it can be shorter and simpler.

ajelenak:
Simpler by avoiding "`data:application/octet-stream;base64:`"? I think that in cases of specs explicit is better than implicit. But avoiding the data URI part will certainly save some bytes.

martindurant:
Updates as requested.

Given the availability of nunjucks (js), liquid (ruby), twig (php) and probably others, I think sticking with jinja2 is fine. It means double-braces, unless all of the above permit setting the variable delimiter as python'd version does.

joshmoore:
> Given the availability of nunjucks (js), liquid (ruby), twig (php) and probably others

Untested but looks active: https://github.com/HubSpot/jinjava

martindurant:
https://github.com/jinja2cpp/Jinja2Cpp

README.md
```diff
@@ -28,7 +29,98 @@ For example, Zarr data in this proposed spec might be represented as:
...
+    "version": 1,
+    "templates": {
+        "u": "server.domain/path",
+        "f": "{{c}}"
+    },
+    "gen": [
+        {
+            "key": "gen_key{{i}}",
+            "url": "http://{{u}}_{{i}}",
+            "offset": "{(i + 1) * 1000}",
```
manzt, line 87:
```suggestion
             "offset": "({{ i }} + 1) * 1000",
```

martindurant: I should be `"{{(i + 1) * 1000}}"` - to evaluate the whole thing within braces as an expression. Your version would evaluate to the string "(0 + 1) * 1000" for i=0.

manzt:
I was able to write a parser in JS to expand the V1 spec: https://github.com/manzt/reference-spec-reader. Thanks for your work on this @martindurant . 

To reiterate a potential concern, `nunjucks` (and I'm guessing other template renders) has a [very clear warning](https://mozilla.github.io/nunjucks/api.html) that evaluating user defined templates is a security vulnerability:

> nunjucks does not sandbox execution so **it is not safe to run user-defined templates or inject user-defined content into template definitions.**

I don't have enough background to know whether this is a concern in this context, but something to be aware of.

martindurant:
The specific concern, at least the example, is about passing strings to the browser, which then evaluates them. In our case, we don't eval strings, we turn strings into other strings with the exception of cast to `int`.

martindurant:
This particular value has three stages:
- arithmetic evaluation based on the value of `i` (no *functions* allowed here except a very limited set that jinja implements)
- forms a string output
- converted to int with a simple cast (either is an int, or error - no arbitrary evaluation).

manzt:
Thanks for walking me through that. This was a misunderstanding on my part (re: which functions are allowed via jinja).

martindurant:
I have specified what the "ref" entries should be, and included the possibility that they are *whole* files (i.e., no offset/length - but we don't need to scan all files to find their lengths).

I thin this can complete the spec, and I can start making the implementation in fsspec as well as a recipe in pangeo-forge ( @rabernat ?).


This whole-file change would account for the grib2 case or small-hdf case, where we suppose there will be a "filter" stage in the zarr loader which takes the bytes, writes them to disc (or BytesIO for HDF), loads and selects data and passes back the numpy buffer. 

This meets the intake-informaticslab use case (especially if we have caching, which could be for the original file or the final array buffer). Of course, specifying this (python) function doesn't happen in this JSON spec, it's either contained in the zarr metadata or even in an intake spec. We could have another convention for those, if we like, such as `grib2`, `hdf:tree/var` or generic `module:func`. I don't intend to implement anything like this quite just yet, but we want it to be possible.

cc @tam203 @arh89

martindurant:
Merging today, unless I hear otherwise. This will be followed by implementation in fsspec's ReferenceFileSystem, and then the pangeo-forge recipe (which will depend on fsspec HEAD).

README.md
```diff
@@ -28,7 +29,100 @@ For example, Zarr data in this proposed spec might be represented as:
...
+        "u": "server.domain/path",
+        "f": "{{c}}"
+    },
+    "gen": [
+        {
+            "key": "gen_key{{i}}",
+            "url": "http://{{u}}_{{i}}",
+            "offset": "{{(i + 1) * 1000}}",
+            "length": "1000",
+            "dimensions":
```
ajelenak, line 91: Somehow `dimensions` seems misplaced to me. What is `i` a dimension of? Perhaps something self-explanatory like "gen_vars", although not as convenient as a single word.

martindurant: Happy to hear other suggestions, but it seems clear enough to me (as described) - in terms of inputs for the cartesian product.

ajelenak: Please go ahead and merge when ready. I did not mean to delay that with my comment.

martindurant:
Implementation: https://github.com/intake/filesystem_spec/pull/568

When that's in, I'll go back fsspec_reference_maker.hdf and edit it accordingly, to be called from a pangeo-forge recipe.

manzt:
Not sure if anyone here is interested, but figured this an appropriate place to share for now (working on a blob post). I have released an JS package on npm that includes a parser for the specification and a valid store implementation for `Zarr.js` in the browser (relies on http(s) protocol).

package: https://github.com/manzt/reference-spec-reader

example: https://6060eed9dc5ef100075be024--vizarr.netlify.app/?source=https://gist.githubusercontent.com/manzt/436fc2966c484205a2c60824f659b412/raw/cdc69f2ce645d953185f10d7552501bfd459dd12/Vanderbilt-Spraggins-Kidney-MxIF.ome.tif.json&channel_axis=0

The example url is a preview of [this PR](https://github.com/hms-dbmi/vizarr/pull/83) in `vizarr`, a viewer for zarr-based images. The reference (v0) was generated via `tifffile` for an OME-TIFF we host on GCS.

rabernat:
Vizarr is so cool! Can we point it at some climate data?

manzt:
@rabernat I'm not sure. I'd have to look at the Zarr (both hierarchy structure and chunk shape for arrays). Currently Vizarr relies on the "multiscales" extension for representing downsampled data, the last two dimensions must be `(y, x)` and non-xy dimensions must have a chunksize of one. The other thing I'm curious about is coordinate system. Happy to chat!

PR #18, opened 2021-03-18
Spec v1 for Ike

martindurant:
For information: templating for the one Ike file reduces the JSON size by 50%.


martindurant:
Thanks @pbranson - I think I was coming to the same conclusion, that I would have to look at the chunks of the dask arrays. I am not sure about where to put the derived coordinates aggregated from the many files, typically "time".

PR #24, opened 2021-05-14
Multi-files

martindurant:


martindurant:
Adds multifile HDF parser and examples

pbranson:
Thanks @martindurant ! I will try test this out on some datasets soon

martindurant:
Note that I have included the ability to output to parquet or zarr here, but fsspec's ReferenceFileSystem can't read those yet. In the non-json case, they will keep the reference data in memory as dataframes, which should save some memory, but probably be just the same speed.

martindurant:
Also note further that I need someone who understands the chunking imposed by xarray open_mfdataset to tell me how to generalise the logic here to more than the example, which has each file at a different time point, so is easy to concatenate in chunks.

pbranson:
Thanks for the tips - from my reading of the code open_mfdataset(chunks={}) will create a separate chunk per file, per variable (it will also ignore underlying chunks on disk, do each dimension of each variable will be a single chunk). However, in this context it doesnt really matter as you are only using mf_opendataset to get the concat dim coordinates (I think). 

I have tried this out here:
https://gist.github.com/pbranson/c65fb24b388f79ac496ef8b6082e5446 

I have misunderstood your comment:
> Note that I have included the ability to output to parquet or zarr here, but fsspec's ReferenceFileSystem can't read those yet. In the non-json case, they will keep the reference data in memory as dataframes, which should save some memory, but probably be just the same speed.

The combined zarr gets created fine, looks to have the V1 spec references in there correctly, however when passing the opened zarr (or a dict it fails) - how do I get the in memory dictionary right for the mapper to work correctly?

martindurant:
> how do I get the in memory dictionary right for the mapper to work correctly?

The dict representation needs to look like the Spec 1 outlines in this repo's README, with keys "version", "templates", "gen" and "refs". A zarr-oriented store is good for putting all the offsets and sizes together, since these integers will pack well, but they need to be unbundled either at load time, or stored in a structure (e.g., dataframe) that allows for fast lookup of a key. The code to do that doesn't exist yet in ReferenceFileSystem, so the JSON representation is the only one that works.

With the data the have, you could unbundle as:
```
dd["refs"] = {
    key: (data if data is not None else [url, offset, size]) for
    (key, data, url, offset, size) in zip(dd["key"], dd["data"], dd["url"], dd["offset"], dd["size"])
}
```
where this unbundling obviously looses the advantage of binary in-memory storage.

pbranson:
Thanks, it seems there is possibly something not quite right with the zarr saved from MultiZarrToZarr as .zgroup is empty (I havent tracked down here). In anycase the .json output works.

> from my reading of the code open_mfdataset(chunks={}) will create a separate chunk per file, per variable (it will also ignore underlying chunks on disk, do each dimension of each variable will be a single chunk). However, in this context it doesnt really matter as you are only using mf_opendataset to get the concat dim coordinates (I think). 

My statement there is incorrect - open_mfdataset respects the underlying chunks when it is opened from zarr. The issue with inconsistent chunks across files is going to be very difficult to solve I think. Some example files where the chunks are inconsistent are March 2016: 
'''
s3://imos-data/IMOS/SRS/SST/ghrsst/L3S-1d/ngt/2016/201603*.nc
'''
MultiZarrToZarr fails when persisting the zarr to disk due to inconsistent chunks. I tried to clear the ds.variable.attrs['encoding'] but that doesnt actually make sense in this context. I guess that given the mapping of the zarr hierarchical keys to array blocks this possibly cant be solved without making the underlying files consistent. Unless we could defined strided chunk reads, which would only work if the chunks aren't compressed. Can you do strided reads from S3?

cc @NickMortimer

martindurant:
It is not a shortcoming of reference-maker/reference-filesystem, but *zarr* required even binning throughout a data variable. I don't think that would make what you ask possible currently. I could conceive of reference-filesystem being able to consolidate multiple chunks into an output (which I presume is what xarray does), but we don't have that kind of logic.

martindurant:
> Can you do strided reads from S3?

No, this is not a thing

martindurant:
xref https://github.com/intake/filesystem_spec/pull/640

PR #29, opened 2021-06-22
Remove explicit import of numcodecs.Shuffle

lsterzinger:
In our meeting last week we discussed removing the direct import `from numcodecs import Zlib, Shuffle` which would fail on `numcodecs < 0.8.0`. 

Now, importing `fsspec-reference-maker.SingleHdf5ToZarr` will not fail if numcodecs is at an older version, but there will still be an import error if a user tries to open a Shuffle compressed data file without the latest version of numcodecs.


martindurant:
perfect

Issue #31, opened 2021-06-22
Creating reference JSONS from netCDF file on Azure Blob hangs


lsterzinger:
I'm trying to use ReferenceMaker on files stored on Azure Blog storage after successfully doing the same for the same files on AWS S3. I'm using the following code:

```python
import fsspec
from fsspec_reference_maker.hdf import SingleHdf5ToZarr
import json

url = 'az://noaa-goes16/ABI-L2-MCMIPF/2020/002/00/OR_ABI-L2-MCMIPF-M6_G16_s20200020000216_e20200020009524_c20200020010031.nc'

so = dict(
         mode="rb", anon=True, default_fill_cache=False, default_cache_type="none",
)

with fsspec.open(url, **so, account_name='goes') as inf:
        h5chunks = SingleHdf5ToZarr(inf, url)
        with open("test.json", 'w') as outf: 
            outf.write(json.dumps(h5chunks.translate()).encode())
```

But this seems to hang forever (or at very least take longer than 20 minutes for a single file, after which I cancelled). This might be an issue with `fsspec` itself, as just opening a file with fsspec and xarray takes 10x as long on Azure as on AWS S3. See my related issue opened here https://github.com/intake/filesystem_spec/issues/681

martindurant:
(solved by adlfs upgrade)

lsterzinger:
Just to update here in case anyone else stumbles upon this issue: 

intake/filesystem_spec#681 did solve the problem of the slow file read times. Running the code above (except using `abfs==0.7.7`,  `goeseuwest` instead of `goes`, and writing the json in mode `'wb'` instead of `'w'`) takes ~40 seconds to generate a JSON. I chose `goeseuwest` since I am running on Microsoft Planetary Computer which is running in the West Europe region, so choosing the data hosted in the same region also serves to speed up access time. 

```python
import fsspec
from fsspec_reference_maker.hdf import SingleHdf5ToZarr
import json

url = 'az://noaa-goes16/ABI-L2-MCMIPF/2020/002/00/OR_ABI-L2-MCMIPF-M6_G16_s20200020000216_e20200020009524_c20200020010031.nc'

so = dict(
        mode="rb", anon=True, default_fill_cache=False, default_cache_type="none",
)

with fsspec.open(url, **so, account_name='goeseuwest') as inf:
        h5chunks = SingleHdf5ToZarr(inf, url)
        with open("test.json", 'wb') as outf: 
[...]

Issue #25, opened 2021-05-14
ideas from H5coro


martindurant:
Two key things we may wish to implement:

- building references on-the-fly, such that the first access to a dataset is slow, but then subsequent accesses use a cached references index; but the user call doesn't change

- merging fetches when contiguous: since we know the full set of start-end offsets we are about to launch for each given URL, we could reduce the number of calls to the remote API and potentially save time.

martindurant:
The second point here is complete and implemented in the fsspec backend.

The first has some POC, e.g. https://github.com/nsidc/earthaccess/pull/278

PR #33, opened 2021-06-24
rewrite MultiZarrToZarr to always use xr.concat

martindurant:
@ lsterzinger @rsignell-usgs - this applies what we discussed, to make to xr.concat explicit and pass a separate set of options to it. If this feels OK, I'll update the docstrings then.
`example_ensamble()` runs fine with this, quite fast (with the "minimal" and "override" arguments)

### DIFF
--- a/fsspec_reference_maker/combine.py
+++ b/fsspec_reference_maker/combine.py
@@ -8,25 +8,25 @@ import fsspec
 import numcodecs
 import xarray as xr
 import zarr
-logging = logging.getLogger('reference-combine')
+logger = logging.getLogger('reference-combine')
 
 
 class MultiZarrToZarr:
 
     def __init__(self, path, remote_protocol,
-                 remote_options=None, xarray_kwargs=None, storage_options=None,
-                 with_mf=True):
+                 remote_options=None, xarray_open_kwargs=None, xarray_concat_args=None,
+                 preprocess=None, storage_options=None):
         """
 
         :param path: a URL containing multiple JSONs
         :param xarray_kwargs:
         :param storage_options:
         """
-        xarray_kwargs = xarray_kwargs or {}
         self.path = path
-        self.with_mf = with_mf
-        self.xr_kwargs = xarray_kwargs
+        self.xr_kwargs = xarray_open_kwargs or {}
+        self.concat_kwargs = xarray_concat_args or {}
         self.storage_options = storage_options or {}
+        self.preprocess = preprocess
         self.remote_protocol = remote_protocol
         self.remote_options = remote_options or {}
 
@@ -36,6 +36,7 @@ class MultiZarrToZarr:
         self.output = self._consolidate(out)
 
         self._write(self.output, outpath)
+        # TODO: return new zarr dataset?
 
     @staticmethod
     def _write(refs, outpath, filetype=None):
@@ -86,7 +87,7 @@ class MultiZarrToZarr:
                 compression="ZSTD"
             )
 
-    def _consolidate(self, mapping, inline_threashold=100, template_count=5):
+    def _consolidate(self, mapping, inline_threshold=100, template_count=5):
         counts = Counter(v[0] for v in mapping.values() if isinstance(v, list))
 
         def letter_sets():
@@ -104,7 +105,7 @@ class MultiZarrToZarr:
 
         out = {}
         for k, v in mapping.items():
-            if isinstance(v, list) and v[2] < inline_threashold:
+            if isinstance(v, list) and v[2] < inline_threshold:
                 v = self.fs.cat_file(v[0], start=v[1], end=v[1] + v[2])
             if isinstance(v, bytes):
                 try:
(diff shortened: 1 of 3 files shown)


rsignell-usgs:
Great news @martindurant!  Which datasets did you try so far?   (I'll try the others)

martindurant:
Only the noaa-nwm-retro-v2.0-pds 10-file example, which is "example_multi" in the hdf module.

lsterzinger:
Can confirm that it works on my 1-day (144 files) of GOES netcdf data in Azure Blob storage. It took the processing time down from 55 minutes to 45 minutes, but only 25 of those minutes were CPU times so I'm not sure what's taking up the rest.

martindurant:
That's the time to create a reference set for each file?

On June 25, 2021 5:53:02 PM EDT, Lucas ***@***.***> wrote:
>Can confirm that it works on my 1-day (144 files) of GOES netcdf data
>in Azure Blob storage. It took the processing time down from 55 minutes
>to 45 minutes, but only 25 of those minutes were CPU times so I'm not
>sure what's taking up the rest.
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/intake/fsspec-reference-maker/pull/33#issuecomment-868851900

-- 

lsterzinger:
That's the time it takes to create a single reference with `MultiZarrToZarr.translate()` from 144 existing reference jsons, I'm using 
```python
mzz = MultiZarrToZarr(
    json_list,
#     "zip://jsons/*.json::combined.zip",
    remote_protocol='az',
    remote_options={
       'account_name' : 'goeseuwest'
    },    
    xarray_open_kwargs={
        'decode_cf' : False,
        'mask_and_scale' : False,
        'decode_times' : False,
        'use_cftime' : False,
        'decode_coords' : False,
    },
    xarray_concat_args={
        "data_vars": "minimal",
        "coords": "minimal",
        "compat": "override",
        "join": "override",
        "combine_attrs": "override",
        "dim": "t"

    }
)

mzz.translate('combined.json')
```

lsterzinger:
I'm thinking this might be something to do with the azure support in fsspec. #31 and intake/filesystem_spec#681 did help my access time, but opening a filesystem from a single reference JSON pointing at 144 netcdf files (`combined.json`, created using the code above) takes 17min using 
```python
fs =  fsspec.filesystem('reference', fo='combined.json', 
                      remote_protocol='az', remote_options={'account_name':'goeseuwest'})
```

Is that expected for a dataset of this size? Each file is ~300 mb.

martindurant:
Just opening the file only needs reading the JSON file. If you run a profile/snakeviz, you'll probably find that rendering the templates is taking a long time - what counts is the number of chunks, not the size of those chunks. Probably you can pass `simple_templates=True` to significantly improve the time. Also, do you have `ujson` installed?

martindurant:
To the earlier question of why it takes so long to do the combination - you should turn on logging in adlfs or referenceFileSystem, to see what remote objects are being fetched. With the "minimal" and "override" options, I would have thought there are not too many, but this might be mistaken. The fact that CPU time is only a fraction of the total time suggests that there is significant contribution from latency, waiting for the remote server.

Also note that the opening of the individual filesystems and datasets as input to the combine class could be done in parallel using dask, much as the `parallel=` option does in [xarray.open_mfdataset](http://xarray.pydata.org/en/stable/generated/xarray.open_mfdataset.html). 


I suggest in any case, that we merge this, so that we can iterate and move on with things such as the docs.

rsignell-usgs:
@martindurant and @lsterzinger , yes, let's merge this and keep on testing.

martindurant:
One optimisation I absultely should have mentioned is `inline_threshold` in `SingleHdf5ToZarr`. It's hard to know what the best value here is, it's a trade-off between small reads and inflating the output JSON, but I've typically been using 100-500bytes.

Finally, I have the comment in #25 that might be relevant: we make no attempt to join fetches with overlapping/contiguous/near-contiguous ranges. This would matter is several small pieces were being loaded from each of the files.

lsterzinger:
@martindurant specifying `simple_templates=True` in `fsspec.filesystem()` didn't seem to do much. 

How do I turn on logging for adlfs? I'm specifying `remote_protocol='az'` instead of `'abfs'`, is there a difference between the two? It seems like they're both using adlfs.

I also ran `fsspec.filesystem()` with a profiler, but I'm 100% confident in my ability to interpret the results. I've attached the profile file here, I was able to visualize it with `snakeviz ./profile`

[profile.zip](https://github.com/intake/fsspec-reference-maker/files/6728511/profile.zip)

martindurant:
"az" and "abfs" are the same

martindurant:
*ALL* of the time is spent in rendering templates with jinja2. Do use `simple_templates=True`, this should avoid jinja entirely, I don't know why it's not making a difference for you.
https://github.com/intake/filesystem_spec/blob/master/fsspec/implementations/reference.py#L229

lsterzinger:
Yeah I am using `simple_templates=True`. Not idea why it's still calling
jinja2 with it specified

lsterzinger:
@martindurant looks like it's calling jinja2 from `_process_references1()` where jinja is used to render regardless 

https://github.com/intake/filesystem_spec/blob/1e5263a3f38af4ba64a5dbaf0414707ed937826d/fsspec/implementations/reference.py#L286

lsterzinger:
At least the call stack says it's called from that function, but it seems like it would only go to jinja if there's `'{{' in u`

martindurant:
Hah! It looks like the following line should simply not be there, it takes time and has no effect. Doubtless this is a bad merge.

lsterzinger:
Okay yeah that makes sense. Should I open a PR to remove it?

lsterzinger:
Removing that line brought the time down from ~20 minutes to just 16 seconds, I think you found the culprit!

martindurant:
Yes please - once you've tried it

On June 28, 2021 5:18:50 PM EDT, Lucas ***@***.***> wrote:
>Okay yeah that makes sense. Should I open a PR to remove it?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/intake/fsspec-reference-maker/pull/33#issuecomment-870050469

-- 

martindurant:
I can think of an even faster way to do it, if the 16s seems too long - if indeed most of the time is still spent in string substitution.

PR #36, opened 2021-06-25
Add docs and github integration

lsterzinger:
I talked with @rsignell-usgs  this morning and he mentioned needing a better place for the examples than as functions inside the codebase.

I set up a quick Sphinx documentation site, where I did an autodoc generation of the two classes (`SingleHdfToZarr` and `MultiZarrToZarr`) which will update when docstrings are updated, as well as an examples page where the code from the two examples are now visible. 

I'm open to feedback or discussion on what you all think should be added/removed/tweaked before this is pulled in. I have a test doc site added here:

https://lsterzinger.github.io/fsspec-reference-maker/

Let me know what you think!


martindurant:
Ok to merge this and iterate? Are you setting up the final docs site?

martindurant:
Merging this now, and we can figure out whatever extra changes may be required afterwards.

lsterzinger:
@martindurant I believe whoever owns the repo needs to enable Github Pages and Github Actions. The action that I included should push to a branch called 'gh-pages'. I believe I had to create an empty branch with that name in my fork in order to get Github Actions to push the docs correctly.

martindurant:
Can you check the log of my push https://github.com/intake/fsspec-reference-maker/runs/2999637344 to see if it did the right thing already?

lsterzinger:
Looks like the workflow was tracking `master` instead of `main`, I opened #42

lsterzinger:
It seems like we're getting the same error here as I was getting in my tests. https://github.com/intake/fsspec-reference-maker/runs/2999729567

The job is failing because there's no `gh-pages` branch to push the docs to. I don't think this can be accomplished via PR, but I think pushing an empty branch to `gh-pages` should be all that's needed. I believe I did:
```
git checkout --orphan gh-pages
git rm -rf .
git commit --allow-empty -m "initial commit"
git push origin gh-pages
```

martindurant:
OK, I did that. Let's see what happens when I (or you) next push something.

lsterzinger:
You should be able to click 're-run jobs' at the top of
https://github.com/intake/fsspec-reference-maker/runs/2999729567

martindurant:
Done. Did it work?
https://github.com/intake/fsspec-reference-maker/runs/2999940010

martindurant:
(if yes, we should update the README to point to the doc page)

lsterzinger:
Looks good! https://intake.github.io/fsspec-reference-maker/

I created #44 to add a badge to the readme

PR #38, opened 2021-06-30
Improved error messages

tedhabermann:
We are testing fsspec-reference-maker on many HDF5 files and most of the break. It is helpful to know more about why they break so I added test-specific error messages.

### DIFF
--- a/fsspec_reference_maker/hdf.py
+++ b/fsspec_reference_maker/hdf.py
@@ -154,10 +154,18 @@ class SingleHdf5ToZarr:
                     f'{h5obj.shape} {h5obj.dtype} {h5obj.nbytes} bytes>')
                 return
 
-            if (h5obj.scaleoffset or h5obj.fletcher32 or
-                    h5obj.compression in ('szip', 'lzf')):
+            #
+            # check for unsupported dataset properties
+            #
+            if h5obj.scaleoffset:
                 raise RuntimeError(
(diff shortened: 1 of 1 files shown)


fsspec_reference_maker/hdf.py
```diff
@@ -154,10 +154,18 @@ def _translator(self, name: str, h5obj: Union[h5py.Dataset, h5py.Group]):
                     f'{h5obj.shape} {h5obj.dtype} {h5obj.nbytes} bytes>')
                 return
 
-            if (h5obj.scaleoffset or h5obj.fletcher32 or
-                    h5obj.compression in ('szip', 'lzf')):
+            #
+            # check for unsupported dataset properties
```
ajelenak, line 158: Change "dataset properties" to "HDF5 filters"

fsspec_reference_maker/hdf.py
```diff
@@ -154,10 +154,18 @@ def _translator(self, name: str, h5obj: Union[h5py.Dataset, h5py.Group]):
...
+            #
+            if h5obj.scaleoffset:
                 raise RuntimeError(
-                    f'{h5obj.name} uses unsupported HDF5 filters')
+                    f'{h5obj.name} uses HDF5 scaleoffset filter - not supported by reference-maker')
+            if h5obj.fletcher32:
```
martindurant, line 163: Have you had a chance to think about or test #35 ?

Issue #32, opened 2021-06-23
missing object_codec for object array


tedhabermann:
I am trying to create references for this [netCDF file](https://ted-habermann.squarespace.com/s/coloradoStations.nc) which is a collection of ~30 xArray dataSets. Each dataSet has 8 HDF5 datasets. This file was written with xArray.

I make it through the root group and the first dataset then I get this error on the second dataset in the group:
Exception has occurred: ValueError
missing object_codec for object array
  File "/Users/metadatagamechanger/GitRepositories/fsspec-reference-maker/fsspec_reference_maker/hdf.py", line 177, in _translator
    za = self._zroot.create_dataset(h5obj.name, shape=h5obj.shape,
  File "/Users/metadatagamechanger/GitRepositories/fsspec-reference-maker/fsspec_reference_maker/hdf.py", line 71, in translate
    self._h5f.visititems(self._translator)
  File "/Users/metadatagamechanger/Library/Mobile Documents/com~apple~CloudDocs/Documents/MetadataGameChanger/ProjectsAndPlans/UNAVCO/CommonContainer/referenceFileSystem/makeReferences.py", line 10, in <module>
    h5t = h5chunks.translate()

On line:             za = self._zroot.create_dataset(h5obj.name, shape=h5obj.shape,
                                            dtype=h5obj.dtype,
                                            chunks=h5obj.chunks or False,
                                            fill_value=h5obj.fillvalue,
                                            compression=compression,
                                            filters=filters,
                                            overwrite=True)

The json so far ends with the reference to the data in the first dataset. The h5ls for the dataset the error comes from is:
/AMC2/Solution           Dataset {7025/7025}
    Attribute: DIMENSION_LIST {1}
        Type:      variable length of
                   object reference
        Data:  (DATASET-1:529616)
    Attribute: _Netcdf4Dimid scalar
        Type:      native int
        Data:  2
    Location:  1:932878
    Links:     1
    Storage:   56200 logical bytes, 112400 allocated bytes, 50.00% utilization
    Type:      variable-length null-terminated UTF-8 string

martindurant:
Did you have strings in the first dataset too? I think we were implicitly assuming regular number grids. However, it should work, if the code in SingleHdf5ToZarr._translator would recognise this encoding and specify numcodecs' "vlen-utf8" - assuming they are the same. Still odd that this shows up in only the second dataset.

martindurant:
I downloaded the file, but doesn't appear to contain anything
```
[...]

Issue #39, opened 2021-06-30
Warnings rather than exits


tedhabermann:
In https://github.com/intake/fsspec-reference-maker/pull/38 I added error messages that were more specific so that we could more easily identify problems with various flavors of HDF5. 

What I would really like is making these RuntimeErrors into warnings  so we could finish processing files instead of exiting. IMHO, the same thing should happen with HDF5 data types that can not be processed, i.e. variable-length strings...

One thing that might work is writing out a -1 for a size of the dataset so that someone could take a look at the json to find errors. Of course, there could be better approaches...

martindurant:
It would be reasonable to warn and continue. Whether to warn or raise could be optional. I would not put special values into the reference output, though, since there's nothing for that in the spec, and it's not clear to me what ReferenceFileSystem should do when loading these. I guess you might end up replacing all values with None?

tedhabermann:
If nothing is added to the output (which is fine), the reference map becomes a map of the data from the file that can be read with Zarr. As there is nothing that says that will be all the data in the file, that is not necessarily a bad thing. it is what it is!

I sort of like writing it into the json file so it is a permanent reminder of what is being missed...

ajelenak:
Perhaps some kind of a value in the JSON that the translation failed for that HDF5 dataset. And even a message describing why it failed (for the record).

martindurant:
I know what you mean, but we have a strict spec for the reference file that we're trying to adhere to (see this repo's readme). Our only use case so far is for loading data with zarr, but this may well not true forever. What I could imagine, is having a specific type of exception that gets raised in ReferenceFileSystem. Any library that doesn't expect it will just raise.

tedhabermann:
I am excited about file maker and am testing it with two large repositories of seismic waveforms and GPS data (IRIS and UNAVC)... We have a wide variety of HDF5 files we are moving to the cloud... My team is getting a little depressed because every time they try to run the code it crashes... Warnings, on the other hand, help us understand the limitations!

martindurant:
Yes, I agree with warnings, so please feel free to edit #38 . Omit the relevant variable for now. It will be good to report back here what the most common problems are so we can deal with them.

martindurant:
Option to warn or ignore now in

PR #40, opened 2021-06-30
Make dict-string codec and apply to HDF5

martindurant:
See #32 

@ajelenak - this doesn't work yet, because the resulting filter object isn't JSON encodable (it contains bytestrings). I assume there's a specific way I'm supposed to handle this.

### DIFF
--- a/fsspec_reference_maker/__init__.py
+++ b/fsspec_reference_maker/__init__.py
@@ -1,0 +1,1 @@
+from .codecs import GRIBCodec, StringDictCodec
--- a/fsspec_reference_maker/codecs.py
+++ b/fsspec_reference_maker/codecs.py
@@ -58,3 +58,4 @@ class GRIBCodec(numcodecs.abc.Codec):
 
 
 numcodecs.register_codec(GRIBCodec, "grib")
+numcodecs.register_codec(StringDictCodec, "dict_string")
--- a/fsspec_reference_maker/codecs.py
+++ b/fsspec_reference_maker/codecs.py
@@ -1,0 +1,60 @@
+import numcodecs.abc
+from numcodecs.compat import ndarray_copy, ensure_contiguous_ndarray
+import numpy as np
+
+
+class StringDictCodec(numcodecs.abc.Codec):
+
+    codec_id = "dict_string"
+
+    def __init__(self, d):
+        """
+
+        :param d: dict(16-byte-string, bytestring)
+        """
+        self.d = d
+
+    def encode(self, buf):
+        # on encode, pass through
+        return buf
+
+    def decode(self, buf, out=None):
+        out = ensure_contiguous_ndarray(out)
+        ids = np.frombuffer(buf, dtype="S16")
+        out[:] = [self.d[i] for i in ids]
+        return out
+
+
+class GRIBCodec(numcodecs.abc.Codec):
+    """
+    Read GRIB stream of bytes by writing to a temp file and calling cfgrib
+    """
+
+    codec_id = 'grib'
+
+    def __init__(self, var):
+        self.var = var
+
+    def encode(self, buf):
+        # on encode, pass through
+        return buf
+
+    def decode(self, buf, out=None):
+        import cfgrib
+        buf = ensure_contiguous_ndarray(buf)
+        fn = tempfile.mktemp(suffix="grib2")
+        buf.tofile(fn)
+
+        # do decode
+        ds = cfgrib.open_file(fn)
+        data = ds.variables[self.var].data
+        if hasattr(data, "build_array"):
+            data = data.build_array()
+
+        if out is not None:
+            return ndarray_copy(data, out)
+        else:
+            return data
+
+
+numcodecs.register_codec(GRIBCodec, "grib")
--- a/fsspec_reference_maker/codecs.py
+++ b/fsspec_reference_maker/codecs.py
@@ -1,3 +1,4 @@
+import base64
 import numcodecs.abc
 from numcodecs.compat import ndarray_copy, ensure_contiguous_ndarray
 import numpy as np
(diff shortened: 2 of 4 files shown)


martindurant:
@ajelenak , I believe this now ought to work, but I get a hard malloc crash (pointer being freed was not allocated). Can you please help me diagnose?

ajelenak:
My first attempt at helping is to recommend a different approach: read the vlen string array and store it in the ReferenceFileSystem JSON. This could be controlled via an option specifying the max byte size of such arrays. Any string array over this max limit would not be stored and this ties in nicely with the similar discussion in #39.

Even the HDF5 code example for reading vlen strings does not use the internal HDF5 struct that is in those "S16" bytes.

martindurant:
Hm, I'm not sure how many data arrays we can possibly inline into the JSON. In the case in hand, the first array would be 63kB with vlen bytes (+expansion with base64), so very doable, but there are many of these in the dataset. The current highest threshold for inlining data I've used is 500bytes.

Even so, I'd really like to know why the code as given in this PR crashes.

ajelenak:
> Hm, I'm not sure how many data arrays we can possibly inline into the JSON. In the case in hand, the first array would be 63kB with vlen bytes (+expansion with base64), so very doable, but there are many of these in the dataset. The current highest threshold for inlining data I've used is 500bytes.

It's the last resort solution, the real goal is to avoid having HDF5 vlen datatypes in files that are to be accessed via ReferenceFileSystem. In this case, I think the source of those vlen HDF5 string datasets is the netCDF library which uses the vlen UTF-8 encoded HDF5 string datatype by default.

> Even so, I'd really like to know why the code as given in this PR crashes.

Can you post the traceback, if there was any? One problem could be the double-use of the `h5f` file-like object in `_make_string_dict()`. `h5f` is already in use by the HDF5 library's virtual file driver (VFD) machinery because h5py opens the HDF5 file with it. An option could be to use the [`H5Dread_chunk()`](http://v1-12-1.hdf5.info/group___h5_d.html#title27) function but those string netCDF variables would need to be chunked (one chunk is enough). @tedhabermann can you create a new file with chunk string variables?

martindurant:
> It's the last resort solution, the real goal is to avoid having HDF5 vlen datatypes in files that are to be accessed via ReferenceFileSystem.

We can't have that goal, as the whole purpose of this effort is to allow access to as wide a range of HDF files as possible, whose provenance we do not control.

> Can you post the traceback, if there was any?

No traceback
```
python(84985,0x109fbadc0) malloc: *** error for object 0x7fdd7b0c2410: pointer being freed was not allocated
python(84985,0x109fbadc0) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

`_make_string_dict` only takes the h5 object and the open file-like (python `io.BufferedReader` for a local file)

martindurant:
@ajelenak : switching to VLenBytes still results in a segfault or malloc crash, apparently during garbage collection. I put in some manual cleanup to the end of `_translator`, but it makes no difference.

ajelenak:
Do you get segfault when these four lines https://github.com/intake/fsspec-reference-maker/blob/642570d7522647a3583b0eae113377020a107731/fsspec_reference_maker/hdf.py#L315-L318 are commented out?

martindurant:
> Do you get segfault when these four lines

Yes - those lines are not even being called as of the latest commit.

tedhabermann:
@ajelenak - a chunk vlen str? You mean write this dataset as chunked? In this case the dataset is written by the xArray to_netcdf function... Not sure how to control that... I'm actually not sure what chunking means in a variable-length string array...

ajelenak:
@tedhabermann That was last week's idea, not needed any more.

ajelenak:
@martindurant What OS did you run this code on? If not Linux, can you try it there?

martindurant:
It was mac - sure, I can try linux

On July 15, 2021 9:44:57 PM EDT, Aleksandar Jelenak ***@***.***> wrote:
***@***.*** What OS did you run this code on? If not Linux, can you
>try it there?
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/intake/fsspec-reference-maker/pull/40#issuecomment-881115990

-- 

martindurant:
Dies:

```
In [9]: out = sin.translate()
2021-07-16 17:46:44,316 - h5-to-zarr - DEBUG - translate -- Translation begins
2021-07-16 17:46:44,327 - h5-to-zarr - DEBUG - _translator -- HDF5 group: /AMC2
2021-07-16 17:46:44,331 - h5-to-zarr - DEBUG - _translator -- HDF5 dataset: /AMC2/Datetime
2021-07-16 17:46:44,332 - h5-to-zarr - DEBUG - _translator -- Created Zarr array: <zarr.core.Array '/AMC2/Datetime' (7025,) int64>
2021-07-16 17:46:44,334 - h5-to-zarr - DEBUG - _translator -- _ARRAY_DIMENSIONS = ['AMC2/Datetime']
2021-07-16 17:46:44,371 - h5-to-zarr - DEBUG - _translator -- HDF5 dataset: /AMC2/Solution
2021-07-16 17:46:44,382 - h5-to-zarr - DEBUG - _translator -- Created Zarr array: <zarr.core.Array '/AMC2/Solution' (7025,) object>
2021-07-16 17:46:44,480 - h5-to-zarr - DEBUG - _translator -- _ARRAY_DIMENSIONS = ['AMC2/Datetime']
Segmentation fault
```

martindurant:
@ajelenak ^ that was my linux (docker) run

martindurant:
Running with the original string dictionary maker, I got extra information, hah
```
In [4]: out = sin.translate()
corrupted double-linked list
Aborted
```

martindurant:
Superseded by #175

PR #43, opened 2021-07-06
Changed "CI" to "Docs"

lsterzinger:
Making this change so that I can create a status badge with the correct name in README.md


martindurant:
Would you like write access to this repo? There shouldn't be any need for me to be involved with simple changes like this.

PR #44, opened 2021-07-06
Added docs badge to README

lsterzinger:
Docs are live at https://intake.github.io/fsspec-reference-maker


martindurant:
I should write some docstrings...
Also, the GRIB and [TIFFfile](https://github.com/cgohlke/tifffile) should be included (the latter is not in this repo)

PR #46, opened 2021-07-07
Added check for HDF5 datasets with type = object

tedhabermann:
This avoids crashes when files with variable-length strings are read. Warnings are posted...

I know there are other HDF5 types that are not supported at present... probably compound datasets... Will add that when I get sample.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,4 @@ dmypy.json
 # Pyre type checker
 .pyre/
 
+.DS_Store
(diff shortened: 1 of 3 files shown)


martindurant:
It would also be cool to figure out why #40 crashes with both the methods I tried. If we can get that to work, we don't need to avoid strings.

martindurant:
Superseded by #175

Issue #52, opened 2021-07-29
"Failed to open Zarr store with consolidated metadata" xarray error with MultiZarrToZarr


lsterzinger:
@rsignell-usgs and I ran into this when working on a notebook together yesterday, xarray seems to have a new warning when it can't open a zarr store with consolidated metadata.

When running `MultiZarrToZarr.translate()`, I get
```
/home/conda/store/578732b5a39dadc3cadc71a29a33e58ded03e8a9d5c888edac6151d80eda2868-pangeo/lib/python3.8/site-packages/fsspec_reference_maker/combine.py:170: 
RuntimeWarning: Failed to open Zarr store with consolidated metadata, falling back to try reading
non-consolidated metadata. This is typically much slower for opening a dataset. 
To silence this warning, consider:
1. Consolidating metadata in this existing store with zarr.consolidate_metadata().
2. Explicitly setting consolidated=False, to avoid trying to read consolidate metadata, or
3. Explicitly setting consolidated=True, to raise an error in this case instead of falling back to try reading non-consolidated metadata.
  dss = [xr.open_dataset(m, engine="zarr", chunks={}, **self.xr_kwargs)
  ```

I'm not sure if we're trying to consolidate metadata or not. If we are, then something isn't working. If we're not, then we might consider adding `consolidated=False` to https://github.com/intake/fsspec-reference-maker/blob/e96969d938572fb883b2acf761506dc11de45a4c/fsspec_reference_maker/combine.py#L170 to suppress the warning for now.

Minimum working code sample and [json files](https://github.com/intake/fsspec-reference-maker/files/6901854/jsons.tar.gz):

```python
import xarray as xr
import fsspec
from glob import glob
from fsspec_reference_maker.combine import MultiZarrToZarr

jsons = sorted(glob("./jsons/*"))

mzz = MultiZarrToZarr(
    jsons,
    remote_protocol='s3',
    remote_options={'anon' : 'True'},   #JSON files  
    xarray_open_kwargs={
        'decode_cf' : False,
        'mask_and_scale' : False,
        'decode_times' : False,
        'use_cftime' : False,
        'drop_variables': ['reference_time', 'crs'],
        'decode_coords' : False
    },
    xarray_concat_args={
#          "data_vars": "minimal",
#          "coords": "minimal",
#          "compat": "override",
        "join": "override",
        "combine_attrs": "override",
        "dim": "time"
    }
)

mzz.translate("test.json")
```

[...]

PR #53, opened 2021-07-29
Added consolidated=False to MultiZarr `open_dataset()` call

lsterzinger:
This suppresses the new warning reported in #52

### DIFF
--- a/fsspec_reference_maker/combine.py
+++ b/fsspec_reference_maker/combine.py
@@ -167,7 +167,7 @@ class MultiZarrToZarr:
             self.fs = fss[0].fs
             mappers = [fs.get_mapper("") for fs in fss]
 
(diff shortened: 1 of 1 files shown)


martindurant:
On reflection, I'd like to include the kwarg *only if* not given, because this class could be used to create aggregate zarr datasets from other normal zarr data - not just out JSON reference based datasets.
```
xrk = self.xr_kwargs.copy()
if consolidated not in xrk:
    xrk['consolidated'] = False
dss = [xr.open_dataset(m, engine="zarr", chunks={}, consolidated=False, **xrk)
               for m in mappers]
```

lsterzinger:
> On reflection, I'd like to include the kwarg only if not given

I updated the PR to pass a copied xr_kwarg variable to `open_dataset` and add `consolidated=False` if not present

martindurant:
OK, cool. I'll merge this and deal with the conflict it'll cause versus #50

PR #54, opened 2021-07-30
h5ds.is_scale

tedhabermann:
The function h5ds.is_scale fails (returns a value < 0) for string type datasets and this prevents writing large fixed-length datasets into HDF5 in a way that can be read by reference-maker. This kind of data can be very useful for writing metadata in csv, xml, json, or text into HDF5 files so that it can be read by Zarr...

Added a check for fixed-length data types (dset.dtype.kind == 'S') to avoid this problem.


martindurant:
Cool, thanks! Do you have a dataset as an example of this kind of thing?

tedhabermann:
Martin - been working on this with Aleks and it turns out that a bug fix in HDF5 1.10.7 fixes this problem with a fix in the HDF5 library. Conda-forge currently supports 1.10.6, the last version prior to this fix being made...

ajelenak:
@tedhabermann It is the conda's h5py that requires HDF5-1.10.6. You can remove h5py from your conda env and use that env's pip command to install h5py. This way h5py will come with its own HDF5-1.12.0 and `h5py.h5ds.is_scale()` should work with your file again.

martindurant:
In the code, we can of course have different behaviour depending on the h5py version (#46 for the warning/bail behaviour conversation).

ajelenak:
This issue is related to the HDF5 library version h5py is using. I learned that the HDF5-1.10.6 version requirement comes from conda-forge channel not conda h5py. There is a request to move to HDF5-1.12.1 but don't know how quickly such things change.

Should this PR still be open? We have now identified the problem and there is a workaround.

ajelenak:
There's a `requirements.txt` file in this repo. Add h5py to it. That should take care of the `pip` installers. I don't see a conda environment config file. To cover those users the config file should specify that h5py be installed via `pip` instead of `conda`.

If the above is acceptable then this PR is not needed. And, in general, the proposed fix is not correct. It just switches off any HDF5 dimension scale detection on string-typed HDF5 datasets.

martindurant:
@tedhabermann , would you like to add a requirement with the correct version spec here, and revert your code change? It makes sense to me in this PR, since it should specifically address your issue.

tedhabermann:
@martindurant @ajelenak - I am flexible and would support whatever you two think is best as you know much more about configuring things than I... It seems unfortunate to break out of the conda forge channel. Hopefully some time soon that will not be necessary (when conda forge upgrades the h5py version).

martindurant:
This package is only installable by pip for the time being, so I don't think it's a problem. If/when I get around to making a feedstock, conda-forge will likely have updated h5py/HDF.

ajelenak:
https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1690 So any help to push this through is welcome.

ajelenak:
From my recent comments, it is obvious I am in favor of resolving this issue by updating install requirements until conda-forge HDF5 library dependency is resolved. If there is no urgency then let's wait on conda-forge.

ajelenak:
One option to resolve this issue is to install h5py from source and thus build it with whatever version of the HDF5 library is available in that conda environment. For example, if possible:

1. `conda install hdf5=1.12.1`
2. `HDF5_DIR=$CONDA_PREFIX pip install --no-binary=h5py h5py`

The second step will need a C compiler.

martindurant:
@ajelenak , I don't think that's reasonable to ask, even of the relatively proficient people that are likely to directly use this package.

ocefpaf:
BTW, conda-forge just started the migration to the latest hdf5 and h5py is already done.

martindurant:
Alright then! I'll close this.

PR #56, opened 2021-08-02
Add ujson requirement to docs build

lsterzinger:
This prevented autodoc from updating docs from docstrings in source


lsterzinger:
It had worked on my local env but it was my mistake not testing the docs on a clean environment

PR #59, opened 2021-08-03
Add support for list-of-dict input to MultiZarrToZarr

lsterzinger:
This PR creates an `fo_list` to pass to `fsspec.filesystem`. Since `fsspec_filesystem` is able to read python dicts directly, if `self.path[0]` is a dict, then fo_list is just a duplicate of that list. 

If `self.path[0]` is not a dict, then it will open the files and read the jsons as dicts, which are added to `fo_list

I tested this by loading some json files and converting them to dictionaries:

```python
json_list = sorted(glob('./jsons/*'))

dict_list = [ujson.load(f) for f in [open(inf) for inf in json_list]]
```
Then running `MultiZarrToZarr` by passing in the dictionary list:
```python
mzz = MultiZarrToZarr(
    dict_list,
    remote_protocol="s3",
    remote_options={'anon':True},
    xarray_open_kwargs={
        "decode_cf" : False,
        "mask_and_scale" : False,
        "decode_times" : False,
        "decode_timedelta" : False,
        "use_cftime" : False,
        "decode_coords" : False
    },
    xarray_concat_args={
        'data_vars' : 'minimal',
        'coords' : 'minimal',
        'compat' : 'override',
        'join' : 'override', 
        'combine_attrs' : 'override',
        'dim' : 't'
    }
)

d = mzz.translate()
print(d)
```

### DIFF
--- a/fsspec_reference_maker/combine.py
+++ b/fsspec_reference_maker/combine.py
@@ -247,24 +247,22 @@ class MultiZarrToZarr:
     def _determine_dims(self):
         logger.debug("open mappers")
 
+        # If self.path is a list of dictionaries, pass them directly to fsspec.filesystem
         if type(self.path[0]) == dict:
-            fss = [
-                fsspec.filesystem(
-                    "reference", fo=d,
-                    remote_protocol=self.remote_protocol,
-                    remote_options=self.remote_options
-                ) for d in self.path
-            ]
-            
+            fo_list = self.path
+        
+        # If self.path is list of files, open the files and load the json as a dictionary
         else:
             with fsspec.open_files(self.path, **self.storage_options) as ofs:
-                fss = [
-                    fsspec.filesystem(
-                        "reference", fo=json.load(of),
-                        remote_protocol=self.remote_protocol,
-                        remote_options=self.remote_options
-                    ) for of in ofs
-                ]
+                fo_list = [json.load(of) for of in ofs]
+
+        fss = [
+            fsspec.filesystem(
+                "reference", fo=fo,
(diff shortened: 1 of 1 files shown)


fsspec_reference_maker/combine.py
```diff
@@ -246,16 +246,25 @@ def _build_output(self, ds, ds0, fss):
...
-                    "reference", fo=json.load(of),
-                    remote_protocol=self.remote_protocol,
-                    remote_options=self.remote_options
-                ) for of in ofs
-            ]
-            self.fs = fss[0].fs
-            mappers = [fs.get_mapper("") for fs in fss]
+
+        # If self.path is a list of dictionaries, pass them directly to fsspec.filesystem
+        if type(self.path[0]) == dict:
```
martindurant, line 251: `isinstance(self.path[0], dict)` - allows for subclasses of dict. `collections.abc.Mapping` may be even better. Also, the path may be a single string with a "*" character (like in `example_ensemble`!), so you would get a TypeError here.

lsterzinger: I think even if `path="*"` the code would work, since `self.path[0]` would return the `*` character as type `str`. I'll update it to use `isinstance` since we want to check for subclasses.
```suggestion
        import collections.abc
        if isinstance(self.path[0], collections.abc.Mapping):
```

martindurant: You can push this - I can't commit for some reason.

lsterzinger: Okay it should be pushed now

Issue #60, opened 2021-08-03
Create tests and automate for new PRs


lsterzinger:
As @martindurant mentioned in #58  we should probably get some tests created with the `example_*` functions and run them automatically with new PRs. I think this can be relatively easy to do with GitHub actions.

lsterzinger:
Thinking about the best way to create tests. I'm starting with the `example_single()` function. I'm thinking that it would be reasonable to generate a "permanent" json file that's added to the repo, generate a dictionary from the example function, and compare it to the dictionary that's read in by `json.load()` on the 'permanent' file.

@martindurant thoughts?

lsterzinger:
Also, the `example_single` function uses a requester pays file. Might be worth switching to a public dataset like GOES

PR #61, opened 2021-08-03
Add reference file generation test

lsterzinger:
I created a unit test with `example_single()` in `hdf.py` using `pytest. I updated the example to use a public dataset instead of a requester_pays dataset. 

The test compares the returned dict to what is written to a JSON I created and committed. I also added a github action that will run the tests for every push and pull request.

The test can be run locally with `pytest tests/reference_tests.py`, but you'll need to install this branch's version of `fsspec-reference-maker` first since I couldn't get relative imports working and the updated `example_single()` is needed.

Once #59 is merged, I'll add a test for MultiZarrToZarr.

Happy to hear suggestions to make this process better! Going to tag #60 here as well


.github/workflows/tests.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Lint with flake8
+        run: |
+          # stop the build if there are Python syntax errors or undefined names
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+      - name: Test with pytest
+        working-directory: ./tests
+        run: |
+          pytest *.py
```
martindurant, line 35: Should probably be `pytest` alone, which will find all the test files in the current tree.

tests/requirements.txt
```diff
@@ -0,0 +1,16 @@
...
+numcodecs
+numpy
+ujson
+pandas
+packaging
+zarr
+xarray
+cx_Freeze
+fastparquet
+s3fs
```
martindurant, line 15: We don't actually need s3fs or adlfs here; also surprised to see wheel, cx_Freeze and packaging.

martindurant (commented):
I believe this won't run until *after* merge, so let's aim to merge quickly, and then we can get it right afterwards. I do have a couple of immediate comments.

tests/requirements.txt
lsterzinger: Without `s3fs`, `fsspec.get_filesystem_class()` raised an error saying 
```
FAILED test.py::test_single - ImportError: Install s3fs to access S3
```

`adlfs` isn't needed, I must have added that by accident. 

`wheel` was added I think because I couldn't `pip install ` the local fsspec-reference-maker package without it I think, but I can't seem to find it in my test logs anymore. It might be fine without it, I'll remove it for now and we'll see what happens

.github/workflows/tests.yml
lsterzinger: When I tried this it didn't seem to find any files, though that may have been an error on my part. I'll change it to `pytest` alone and we'll see what happens

lsterzinger:
```suggestion
          pytest
```

tests/requirements.txt
lsterzinger: You're right I think for `cx_Freeze` and `packaging`, I removed those too. They were added when I used `pipreqs` to determine all the dependencies

lsterzinger:
Looks like the checks are failing without specifying `pytest *.py`, I'm seeing 
```
============================= test session starts ==============================
platform linux -- Python 3.7.11, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /home/runner/work/fsspec-reference-maker/fsspec-reference-maker
collected 0 items

============================ no tests ran in 0.00s =============================
Error: Process completed with exit code 5.
```
In the Github Actions log

martindurant:
The tests should not be in the source files, but collected into test files, usually in a deirectory called "tests" and with file names "test_*.py". They should import ffspec_reference_maker.

lsterzinger:
Perfect, renaming the file to `test_single_hdf.py` seems to have made `pytest` see the file properly. I also moved the test code into the test file itself.

martindurant:
There seems to be some repetition of the test code and JSON file?

lsterzinger:
> There seems to be some repetition of the test code and JSON file?
I'm curious what you mean... I created a JSON that will be compared to the output from the test code. If something breaks the test code, the dictionary created from it should be different than the JSON that's committed to the repo

lsterzinger:
I added a MultiZarr test, but the test is failing because it's missing `jninja2`. We could either add it as a dependency or give `simple_templates=True` somewhere, which I think bypasses jninja

martindurant:
I recommend doing both!

lsterzinger:
Now it's failing because it needs dask in `mzz.translate()`. Does that sound right? I wasn't sure if things should be able to proceed effectively without dask or not.

Also, I've made a ton of tiny commits to this PR. Happy to squash things down if you'd like to keep things cleaner

martindurant:
Oh indeed, we have `chunks={}`, which is xarray-speak for "use dask".

martindurant:
>  Happy to squash things down if you'd like to keep things cleaner

Don't worry about it, github can squash too.

lsterzinger:
Cool, that seems to cleared up all the failed tests. I see all tests (python 3.7, 3.8, 3.9) succeeding now

lsterzinger:
> I still see single_example.json in two places, inside tests/ and tests/example_jsons , and tests/test_hdf.py repeats code in tests/test.py and the example_* functions.

Ah my bad, I thought I committed the deletion. The code in `test_hdf.py` is self-contained and no longer requires of the `example_*` functions from source. I can delete the ones I used from the source (example_single, example_multi, and example_ensemble) if you'd like.

martindurant:
... and now need unconflicting :)

PR #63, opened 2021-08-05
Added option to turn off templates and allowed for infinite lettersets

lsterzinger:
`MultiZarrToZarr.translate()` now accepts a `template_count=x` kwarg. Default is 5, change to `None` to disable templates entirely.

`letter_sets()` has also been updated to generate an infinite amount of template identifiers to better support large MultiZarrToZarr datasets


martindurant:
Add a test? It'd be interesting to see the difference in file size, and load time.

lsterzinger:
The `multizarr_example.json` is 22 KB with templates, and when I turn off templates it's 43 KB, so nearly double the size.

Loading in the dataset with `fsspec.filesystem()` and `xr,open_dataset()` took 728 ms for the file with references, and 735 ms for the one without references, from my local machine.

martindurant:
Interesting numbers, thanks. Doubtless it depends greatly on the specifics of a dataset. Binary representations with internal compression (parquet, zarr) should do very well with the repeated strings when avoiding templates.

PR #67, opened 2021-08-20
Require s3fs >=2021.08.0 for aiobotocore error

lsterzinger:


lsterzinger:
Whoops I had meant to commit this directly instead of a PR, but I updated `tests/requirements.txt` to use `s3fs>=2021.08.0` to fix the `aiobotocore` error that was failing the tests. (also mentioned here https://github.com/dask/s3fs/issues/514)

PR #70, opened 2021-08-26
Use xarray to concat time coords, keeping original units

martindurant:


lsterzinger:
For a big change like this, it might be best for the tests to alter the example `.json` file that the code dict output is being compared to. Or, like you mentioned a while back, comparing specific bits of a dataset (or something else that doesn't compare dicts directly). Maybe check coordinate arrays (make sure time is still OK), and a few samples of data? I might have some time this week to work on that.

martindurant:
@lsterzinger - that is exactly what's I'm doing; and finding out that something is going wrong! 

I am following this up before tacking @rabernat 's suggestion to use xarray more directly in computing the concat coord. I'm not sure if that's tractable, because we explicitly do NOT want to open all the datasets with xarray - but maybe we can open just the parts of the coordinate arrays and concat those.

rabernat:
> ecause we explicitly do NOT want to open all the datasets with xarray - but maybe we can open just the parts of the coordinate arrays and concat those.

That's exactly what is done in https://github.com/pangeo-forge/pangeo-forge-recipes/blob/a23cb8266171890b581859576ecd7188808605e3/pangeo_forge_recipes/recipes/xarray_zarr.py#L587-L596

martindurant:
Yes, agreed - I have not yet had chance to try that.

lsterzinger:
> @lsterzinger - that is exactly what's I'm doing; and finding out that something is going wrong!

Gotcha, I didn't read the commits clearly enough... :laughing: 

I must have looked at an old failed test where it had failed on the json/dict comparison.

rabernat:
Some comments to help with the implementation I am suggesting

First, let me explain how xarray handles encoding. To understand this, it is useful to understand an xarray [Variable](http://xarray.pydata.org/en/stable/generated/xarray.Variable.html). Variables are not part of the public API but are very useful internally and for lower-level tools. The important part to remember is that a Variable has four parts: `dims`, `data`, `attributes`, and `encoding`. The `data` is a raw np.array-like object. `attributes` and `encoding` are dicts.

A variable can be in either an _encoded_ or _decoded_ state. To actually analyze the data, we need it _decoded_. When the data are encoded, the information about encoding lives in `attributes`. When the data are decode, the information about encoding lives in `encoding`. Xarray has functions that we can use to move a variable between the two states.

Now let's walk through how the linked [pangeo forge code](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/a23cb8266171890b581859576ecd7188808605e3/pangeo_forge_recipes/recipes/xarray_zarr.py#L587-L596) works

```python
# we already have an in-memory, decoded Xarray dataset called ds_chunk
# and are looping through this as
# for vname, var_coded in ds_chunk.variables.items():
# so now var_coded is an xr.Variable
# zgroup is the _target_ zarr group, where we want to write the data
# with consistent encoding

# get the array we want to write to
zarr_array = zgroup[vname]

# update the variable encoding to match the target
var_coded.encoding.update(zarr_array.attrs)

# just delete all attributes from the var;
# they are not used anyway, and there can be conflicts
# related to xarray.coding.variables.safe_setitem
var_coded.attrs = {}

# here is where the work happens; we convert the in-memory decoded variable
# to an encoded variable
var = xr.backends.zarr.encode_zarr_variable(var_coded)

# get as numpy data (e.g. trigger dask computation, coerce lazy arrays to in memory)
data = np.asarray(var.data)

# actually write the encoded data
zarr_array[zarr_region] = data
```

This could easily be used in fsspec-reference-maker without starting from xarray data. you just have to create an xarray.Variable object for the data in question, using the zarr attrs, which is very cheap.

martindurant:
Thanks for describing the difference between Variable and dataArrays - I was certainly confused, since h5py has Variables too.

On September 3, 2021 4:11:56 PM EDT, Ryan Abernathey ***@***.***> wrote:
>Some comments to help with the implementation I am suggesting
>
>First, let me explain how xarray handles encoding. To understand this, it is useful to understand an xarray [Variable](http://xarray.pydata.org/en/stable/generated/xarray.Variable.html). Variables are not part of the public API but are very useful internally and for lower-level tools. The important part to remember is that a Variable has four parts: `dims`, `data`, `attributes`, and `encoding`. The `data` is a raw np.array-like object. `attributes` and `encoding` are dicts.
>
>A variable can be in either an _encoded_ or _decoded_ state. To actually analyze the data, we need it _decoded_. When the data are encoded, the information about encoding lives in `attributes`. When the data are decode, the information about encoding lives in `encoding`. Xarray has functions that we can use to move a variable between the two states.
>
>Now let's walk through how the linked [pangeo forge code](https://github.com/pangeo-forge/pangeo-forge-recipes/blob/a23cb8266171890b581859576ecd7188808605e3/pangeo_forge_recipes/recipes/xarray_zarr.py#L587-L596) works
>
>```python
># we already have an in-memory, decoded Xarray dataset called ds_chunk
># and are looping through this as
># for vname, var_coded in ds_chunk.variables.items():
># so now var_coded is an xr.Variable
># zgroup is the _target_ zarr group, where we want to write the data
># with consistent encoding
>
># get the array we want to write to
>zarr_array = zgroup[vname]
>
># update the variable encoding to match the target
>var_coded.encoding.update(zarr_array.attrs)
>
># just delete all attributes from the var;
># they are not used anyway, and there can be conflicts
># related to xarray.coding.variables.safe_setitem
>var_coded.attrs = {}
>
># here is where the work happens; we convert the in-memory decoded variable
># to an encoded variable
>var = xr.backends.zarr.encode_zarr_variable(var_coded)
>
># get as numpy data (e.g. trigger dask computation, coerce lazy arrays to in memory)
>data = np.asarray(var.data)
>
># actually write the encoded data
>zarr_array[zarr_region] = data
>```
>
>This could easily be used in fsspec-reference-maker without starting from xarray data. you just have to create an xarray.Variable object for the data in question, using the zarr attrs, which is very cheap.
>
>-- 
>You are receiving this because you authored the thread.
>https://github.com/intake/fsspec-reference-maker/pull/70#issuecomment-912784677
-- 

PR #71, opened 2021-09-02
NOT

martindurant:


rsignell-usgs:
Ship it!  https://nbviewer.jupyter.org/gist/rsignell-usgs/587c15d3d9fb8c1e9c41adc7305e41cb

PR #73, opened 2021-09-03
Also inline non-concat coords

martindurant:
@rsignell-usgs - appears to do the right thing for `time` as well as `valid_time` for HRRR.


lsterzinger:
I added the fix from #75 as a temporary measure until the changes @rabernat mentioned in #70 are implemented

fsspec_reference_maker/combine.py
```diff
@@ -316,3 +278,53 @@ def _determine_dims(self):
...
+            else:
+                accum.append(zz)
+        except Exception as e:
+            ex = e
+            accum.append(zz[...].copy())
+    attr = dict(z[accum_dim].attrs)
+    if times:
+        accum = [np.array(a, dtype="M8") for a in accum]
+        attr.pop('units')
+        attr.pop('calendar')
```
martindurant, line 321: Didn't you say that we should remove "calendar" in every case, since we didn't necessary get to `times = True` after setting it in the block above?

lsterzinger: Sure, but `.pop()` raises `KeyError` if there's no "calendar" attribute. So maybe something like this

```python
for key in ('units', 'calendar'):
  try:
    attr.pop(key)
  except KeyError:
    pass

Comment (martindurant) [reply fsspec_reference_maker/combine.py:321]: `pop("calendar", None)`

File: fsspec_reference_maker/combine.py [hunk 2]
```diff
@@ -316,3 +278,54 @@ def _determine_dims(self):
...
+            else:
+                accum.append(zz)
+        except Exception as e:
+            ex = e
+            accum.append(zz[...].copy())
+    attr = dict(z[accum_dim].attrs)
+    if times:
+        accum = [np.array(a, dtype="M8") for a in accum]
+
+    attr.pop('units', None)
```
martindurant, line 321: Actually, there's a problem here. "units" can be used for all kinds of legitimate things that are nothing to do with cftime; whereas calendar is much less likely. I think I would push this line back into the if block to be safe.

lsterzinger: Ah makes sense! I'll move that back under the if block. I've also been playing around with the method @rabernat suggested, if I get it PR-worthy would you like me to push to this PR or open a new one?

martindurant: Let's make that a new one.

PR #80, opened 2021-09-24
Replace empty h5py empty object with empty string.

tinaok:
Json can not handle `h5py.Empty` object, thus if a NetCDF file contains that, `SingleHdf5ToZarr.translate` fails.   We change that object to `""` as Xarray does.  Work done thanks to  @keewis

### DIFF
--- a/fsspec_reference_maker/hdf.py
+++ b/fsspec_reference_maker/hdf.py
@@ -132,6 +132,8 @@ class SingleHdf5ToZarr:
                         v = v.tolist()
                 else:
                     v = v.tolist()
+            elif isinstance(v, h5py._hl.base.Empty):
+                v = ""
             if v == 'DIMENSION_SCALE':
                 continue
             try:


martindurant:
Do you have a test dataset for this? Mostly, I'm wondering what you're working on.

tinaok:
Thank you for merging.  The dataset we use are NEMO-ICE(https://www.nemo-ocean.eu) output created with XIOS (https://forge.ipsl.jussieu.fr/ioserver) in netcdf format.  It has one data variable with an empty units attribute.  We saw the same problem in this MARS3D data sets too.  https://www.ifremer.fr/sextant_doc/pages/DATARMOR.html#/metadata/55ceb588-abc4-4101-9d7e-51f411673b54 (sorry it is in French, but you can click 'download'  icons there, feel free to mail me if you need any more informations)

martindurant:
Do you plan on processing these data sets with pangeo-forge? 
If not, we should maybe start a registry or catalogue in this repo of datasets that have been referenced, as examples for others wondering what is possible with reference-maker.

PR #101, opened 2021-11-02
Using JSON file instead of intake

jkellndorfer:
Changing the cell calling intake with a direct call to mapping via the JSON file describing the zarr format which resides on the AWS registry of open data repository where the data set resides as well. Users will need to change to their working kernel.

### DIFF
--- a/examples/earthbigdata.ipynb
+++ b/examples/earthbigdata.ipynb
@@ -42,17 +42,18 @@
    "id": "a13456f4-35ce-4254-a172-ca26b320661d",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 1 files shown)


martindurant:
@rsignell-usgs , this is essentially what you ended up doing, right?

rsignell-usgs:
@martindurant , yes, I ended up using this blob from @jkellndorfer  instead of the intake catalog. 
It would be nice to have the intake catalog working.   My understanding is that the problem was:
* `.open_data` should have been `.open_catalog`
* the intake catalog used assumes some version of intake that has not yet been released that infers things that previous versions didn't

Yes?

martindurant:
I'll get it working (including releases), but merge this for now.

PR #109, opened 2021-12-17
Remote write for combine

martindurant:
Plus some basic testing


martindurant:
@rsignell-usgs , a very clear example (if you look through the last couple of commits) of how time units completely mess things up.

PR #116, opened 2022-01-10
document the option of manually loading and writing the json metadata

keewis:
`MultiZarrToZarr` uses `fsspec.open_files` to open all files at once. Even though this makes sense for remote filesystems, it fails with a lot of small files on a local filesystem: opening a lot of files at once will exceed the limit for open file descriptors. It is still possible to use `MultiZarrToZarr` in this case because `path` can be passed the pre-loaded metadata. This, however, is currently undocumented.

Writing to a file object is also not possible (I was trying to use `fsspec`'s compression support), but `translate` will return the metadata if it doesn't receive a value for `outpath`. While this is documented, I didn't immediately understand how to use that so I hope this might make it a bit clearer.

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -65,6 +65,12 @@ class MultiZarrToZarr:
 
         storage_options : dict
             When using an fsspec URL to write to remote
+
+        Returns
+        -------
+        dict or None
+            If outpath was given, the metadata is encoded as json and
+            written to that file. If not, it is returned directly.
         """
         self._open_mappers()
         ds, ds0, fss = self._determine_dims()
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -18,9 +18,11 @@ class MultiZarrToZarr:
 
     Parameters
     ----------
-    path: string or list of strings
-        List of JSON paths or a URL containing multiple JSONs
-    
+    path: string, list of strings or list of dict
+        List of JSON paths or a URL containing multiple JSONs. If a
+        list of mappings, this is the json-decoded content of the
+        files.
+
     remote_protocol: string
         Protocol used to access remote files (e.g. 's3', 'az', etc)
 
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -21,7 +21,7 @@ class MultiZarrToZarr:
     path: string, list of strings or list of dict
         List of JSON paths or a URL containing multiple JSONs. If a
         list of dicts, this is the json-decoded content of the
-        files (such as the output of many calls to``kerchunk.hdf.SingleHdf5ToZarr()``.
+        files (such as the output of many calls to``kerchunk.hdf.SingleHdf5ToZarr()``).
 
     remote_protocol: string
         Protocol used to access remote files (e.g. 's3', 'az', etc)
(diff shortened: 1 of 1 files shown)


keewis:
It might be better to change the behavior of both, though: `MultiZarrToZarr.__init__` could read the files in batches (if `fsspec.open_files` supports that), and I think `translate` should be split into `translate` and a new `write_metadata` function: changing the return value based on the input parameter feels a bit strange to me.

kerchunk/combine.py
```diff
@@ -18,9 +18,11 @@ class MultiZarrToZarr:
 
     Parameters
     ----------
-    path: string or list of strings
-        List of JSON paths or a URL containing multiple JSONs
-    
+    path: string, list of strings or list of dict
+        List of JSON paths or a URL containing multiple JSONs. If a
+        list of mappings, this is the json-decoded content of the
+        files.
```
lsterzinger, line 24:
```suggestion
    path: string, list of strings or list of dict
        List of JSON paths or a URL containing multiple JSONs. If a
        list of dicts, this is the json-decoded content of the
        files (such as the output of many calls to``kerchunk.hdf.SingleHdf5ToZarr()``.
```

lsterzinger (commented):
Thanks for helping clarify the documentation! Recommended one small change to hopefully make things more clear.

martindurant:
> MultiZarrToZarr.__init__ could read the files in batches (if fsspec.open_files supports that)

fsspec.open_files returns an OpenFiles instance, which when used in a `with` context to open all of them at once. However, the instance is also a list, and you can slice or iterate over it, to get the OpenFile (no s!) instances. The actual real file is only opened when these are put in a context:

```
ofs = fsspec.open_files(...)
batch = ofs[:10]
for of in batch:
    with of as f:
        # use f
```

Unfortunately, `batch` is not an OpenFiles instance, but a normal list here (could be fixed), so you can't do `with batch` and open all the files in the subset at once in a single context.

martindurant:
The following in fsspec appears to fix the comment above, if we want it
```diff
--- a/fsspec/core.py
+++ b/fsspec/core.py
@@ -204,6 +204,12 @@ class OpenFiles(list):
                     break
         [s.__exit__(*args) for s in self]

+    def __getitem__(self, item):
+        out = super().__getitem__(item)
+        if isinstance(item, slice):
+            return OpenFiles(out, mode=self.mode, fs=self.fs)
+        return out
+
     def __repr__(self):
         return "<List of %s OpenFile instances>" % len(self)
```

kerchunk/combine.py
```diff
@@ -18,9 +18,11 @@ class MultiZarrToZarr:
 
     Parameters
     ----------
-    path: string or list of strings
-        List of JSON paths or a URL containing multiple JSONs
-    
+    path: string, list of strings or list of dict
+        List of JSON paths or a URL containing multiple JSONs. If a
+        list of dicts, this is the json-decoded content of the
+        files (such as the output of many calls to``kerchunk.hdf.SingleHdf5ToZarr()``.
```
keewis, line 24:
```suggestion
        files (such as the output of many calls to``kerchunk.hdf.SingleHdf5ToZarr()``).
```

PR #120, opened 2022-01-28
Add FITS

martindurant:
Extracting from #83 

Also added ASCIITable support. Will shortly add BinTable support (which is easier!).


martindurant:
(test fails because numcodecs hasn't been released yet)

martindurant:
This will do for me for now - it's certainly a good start!

PR #131, opened 2022-03-02
Netcdf3

martindurant:


kerchunk/netCDF3.py
```diff
@@ -0,0 +1,114 @@
...
+
+class netcdf_recording_file(netcdf_file):
+
+    def __init__(self, filename, *args, storage_options=None, **kwargs):
+        assert kwargs.pop("mmap", False) is False
+        assert kwargs.pop("mode", "r") == "r"
+        assert kwargs.pop("maskandscale", False) is False
+        self.chunks = {}
+        with fsspec.open(filename, **(storage_options or {})) as fp:
+            super().__init__(fp, *args, mmap=False, mode="r", maskandscale=False, **kwargs)
```
rabernat, line 17: AFAICT, any time you use `mmap=False`, you will trigger the memory spike reported in https://github.com/pangeo-forge/pangeo-forge-recipes/issues/361. I think scipy basically load the entire dataset into memory. So I'm not sure that this approach to constructing references is as lazy as one would hope.

rabernat: Here is a good test file: https://cluster.klima.uni-bremen.de/~fmaussion/teaching/climate/dask_exps/HiRes_Hourly_Surf/ERA5_HiRes_Hourly_tp_2000_01.nc

martindurant: Now produces:
```
<xarray.Dataset>
Dimensions:    (latitude: 721, longitude: 1440, time: 744)
Coordinates:
  * latitude   (latitude) float32 90.0 89.75 89.5 89.25 ... -89.5 -89.75 -90.0
  * longitude  (longitude) float32 0.0 0.25 0.5 0.75 ... 359.0 359.2 359.5 359.8
  * time       (time) datetime64[ns] 2000-01-01 ... 2000-01-31T23:00:00
Data variables:
    tp         (time, latitude, longitude) float32 ...
```

Note that this file doesn't have an append-only dimension,  which doens't work yet, and you get exactly one chunk per array; but subchunking will be easy to implement.

martindurant:
I intend to merge this useful but rather incomplete implementation. Subchunking and the append dimension will wait.

PR #137, opened 2022-03-25
Fix netcdf datetime handling

### LINKED ISSUE #115: Incorrect times after `
```python

### PR DESCRIPTION
**Pull request by gadomski**
Fixes netcdf datetime handling by removing the fill value if the dataset is a netcdf datetime. NetCDF datetimes are detected using the same heuristic that xarray uses: https://github.com/pydata/xarray/blob/f8bae5974ee2c3f67346298da12621af4cae8cf8/xarray/coding/times.py#L670.

## Related issues
- Closes #115 

## Additional changes
- Adds more required dependencies to `requirements.txt`
- Adds a `requirements-dev.txt` with development dependencies, for developers that aren't using conda

### DIFF
--- a/kerchunk/hdf.py
+++ b/kerchunk/hdf.py
@@ -173,6 +173,8 @@ class SingleHdf5ToZarr:
                 compression = None
             if h5obj.dtype.kind in "UVS":
                 fill = h5obj.fillvalue or " "
+            elif _is_netcdf_datetime(h5obj):
+                fill = None
             else:
                 fill = h5obj.fillvalue
             
@@ -302,3 +304,12 @@ class SingleHdf5ToZarr:
                     [a // b for a, b in zip(blob.chunk_offset, chunk_size)])
                 stinfo[key] = {'offset': blob.byte_offset, 'size': blob.size}
             return stinfo
+
+
+def _is_netcdf_datetime(dataset: h5py.Dataset):
+    units = dataset.attrs.get("units")
+    if isinstance(units, bytes):
+        units = units.decode("utf-8")
+    # This is the same heuristic used by xarray
+    # https://github.com/pydata/xarray/blob/f8bae5974ee2c3f67346298da12621af4cae8cf8/xarray/coding/times.py#L670
+    return units and "since" in units
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,5 @@
 fsspec
+numcodecs
+numpy
 ujson
+zarr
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: requirements.txt [hunk 1]
```diff
@@ -1,2 +1,9 @@
+cftime
```
martindurant, line 1: Please do not add all thee requirements. If we wanted to kerchunk FITS files, for example, we wouldn't need all these. You could add them to optional install selectors, to install like `pip install kerchunk[cdf,grib]`...

kerchunk/hdf.py
```diff
@@ -302,3 +304,10 @@ def _storage_info(self, dset: h5py.Dataset) -> dict:
                     [a // b for a, b in zip(blob.chunk_offset, chunk_size)])
                 stinfo[key] = {'offset': blob.byte_offset, 'size': blob.size}
             return stinfo
+
+
+def _is_netcdf_datetime(dataset: h5py.Dataset):
+    attrs = dataset.attrs
+    # This is the same heuristic used by xarray
+    # https://github.com/pydata/xarray/blob/f8bae5974ee2c3f67346298da12621af4cae8cf8/xarray/coding/times.py#L670
+    return "units" in attrs and "since" in attrs["units"]
```
martindurant, line 313: Should be `attrs.get("units", "")`, since the attribute might not exist.

martindurant (commented):
This seems like a reasonable thing to do for this very specific case. It is interesting that xarray uses this logic without invoking cftime directly.

Note that this case is already dealt with at combine time, by specifying that the given input data types are to be interpreted using cftime. This is done by adding, for example `{"time": "cf:time"}` to the coordinate mapper, which says "calculate the coordinate values of 'time' by applying cftime to the data variable called 'time'"

kerchunk/hdf.py
gadomski: Shouldn't the initial `"units" in attrs` handle that case?

martindurant: Sorry, I'm talking rubbish, ignore me :)

gadomski: No worries :-)

requirements.txt
```diff
@@ -1,2 +1,6 @@
+cfgrib
```
martindurant, line 1: cfgrib is also strictly optional

gadomski: Fixed.

martindurant:
👍

PR #142, opened 2022-04-02
Update combine.py

RichardScottOZ:
concat_dims could be None - should that be self.concat_dims?


martindurant:
This would have saved @rsignell-usgs a step!

PR #144, opened 2022-04-06
Missing dims and tifffile wrapper

martindurant:


martindurant:
(well, I suppose tifffile can write some stuff for us...)

rsignell-usgs:
Yes I do!
```
import fsspec
fs = fsspec.filesystem('s3', anon=True, client_kwargs={'endpoint_url': 'https://mghp.osn.xsede.org'})
flist = fs.ls('/rsignellbucket1/lcmap/cog/')
print(flist)
```
results in:
```
['rsignellbucket1/lcmap/cog/LCMAP_CU_2019_V12_LCPRI.tif',
 'rsignellbucket1/lcmap/cog/LCMAP_CU_2020_V12_LCPRI.tif']
```
that will work, right?

martindurant:
I'd rather not have to reach to the internet for tests, though - and these are far too big to put in the repo. We could create random data, I suppse.

rsignell-usgs:
We could cut out a small subset (2048x2048, for example)

martindurant:
OK, I think this is useful as it is.

PR #149, opened 2022-04-13
Update hdf.py

RichardScottOZ:
possible fill = None fix for now ??

Noticed this when using switching to using latest kerchunk in a container on AWS.


martindurant:
I think this is right. Do you experience the expected exception without this?

RichardScottOZ:
Yes, I did.

Issue #145, opened 2022-04-07
`mzz.translate()` much slower in 0.0.6


lsterzinger:
I'm helping a colleague use kerchunk to access GOES-16 satellite data on AWS. At first I was running Kerchunk 0.0.5 and combining 288 references took a few seconds, nothing special.

We updated to 0.0.6 because there was an issue with some of the coordinates and I figured we'd benefit from the MultiZarrToZarr update. Now it I let it run for over 10-15 minutes and it never finished. Not quite sure what's going on, or if I'm missing some important optimization somewhere.

Here is the notebook: https://nbviewer.org/gist/lsterzinger/37c035542f7fac8f7e02a7c736877b60

martindurant:
It's totally possible that the new MZZ is much slower than previous, I didn't make a great effort to optimise for performance. However, it's also quite likely that there's something else going on, for example maybe many <500byte chunks are being inlined, and fetching there is currently a serial process, so many such chunks really add up.

The thing to do would be to try with say 3 inputs, and then 10, and see the time it takes, and also of course to profile (snakeviz, and lprun on MultiZarrToZarr.second_pass), so we know where time is being spent.

lsterzinger:
3 files ~ 30 seconds
10 files ~ 1 min 30 seconds

So this appears to scale linearly.

Running a profile on `mzz.second_pass()` [(second_pass.zip)](https://github.com/fsspec/kerchunk/files/8453421/second_pass.zip) shows that nearly all of the time is spent on `~:0(<method 'acquire' of '_thread.lock' objects>)` which doesn't really tell us anything... I've run into this before and I don't really know how to investigate further

![image](https://user-images.githubusercontent.com/6716906/162475441-a507235d-4555-46d4-a45d-e95de7a64873.png)

martindurant:
Oh, I see: 2071 calls, at about 43ms each - sounds like pure latency, so we can easily do 100 of these caslls simultaneously.

lsterzinger:
So this is actually with  an `inline_threshold=0`... reference.py:246(cat) is being called 2071 times

martindurant:
MultiZarrToZarr.second_pass only calls `cat` [here](https://github.com/fsspec/kerchunk/blob/main/kerchunk/combine.py#L369), which only happens for a file-size smaller than the threshold. Are you sure you are passing it correctly? 

In any case, the following, I think is enough to get one order of magnitude of concurrency:
```diff
--- a/kerchunk/combine.py
[...]

PR #154, opened 2022-05-04
Faster MultiZarrToZarr performance

lsterzinger:
Implements @martindurant's fix from #145


martindurant:
I'm pretty sure it can be better than this - because each of the "filesystems" in the loop is probably running on the same implementation (s3, http...) - but it will do for now!

martindurant:
Looks like the doctest run environment has gone stale

PR #175, opened 2022-06-14
copy with compound VLEN strings

martindurant:


martindurant:
@sharkinsspatial , this code is enough to successfully process "s3://gedi-test/GEDI04_A_2019108045816_O01962_01_T01066_02_002_02_V002.h5", and be able to open it with zarr (not xarray, since this is not netCDF compliant!). However, the strings in the embedded tables show their obfuscated 16-byte string IDs, not the real values. This is a start! I can write a codec which recodes from those IDs to the actual values, which should be applied when there are fewer than some configurable number of values, which will work well for all those "sqrt" strings, for instance.

Issue #176, opened 2022-06-14
"TypeError: reject_bytes is on.." when utilising postproces in MultiZarrtoZarr


peterm790:
Postprocess in [MultiZarrtoZarr](https://github.com/fsspec/kerchunk/blob/main/kerchunk/combine.py) is a useful way to modify attributes such as that suggested in https://github.com/fsspec/kerchunk/issues/173 

However at present the following will work to create a combined json file but will not work when writing to json: 

      import zarr 
      def modify_fill_value(out):
          out_zarr = zarr.open(out)
          out_zarr.y.fill_value = 10
          return out
      
      def postprocess(out):
          out = modify_fill_value(out)
          return out

instead the following is necessary:

      import base64
      
      def consolidate(out):
          for k, v in out.items():
              if isinstance(v, bytes):
                  try:
                      # easiest way to test if data is ascii
                      out[k] = v.decode('ascii')
                  except UnicodeDecodeError:
                      out[k] = (b"base64:" + base64.b64encode(v)).decode()
              else:
                  out[k] = v
          return out
      
      import zarr
      
      def modify_fill_value(out):
          out_zarr = zarr.open(out)
          out_zarr.y.fill_value = 10
          return out
      
      def postprocess(out):
          out = modify_fill_value(out)
          out = consolidate(out)
          return out 

a reproducible notebook using post-process is available here https://gist.github.com/peterm790/4b1632cfea9f76cd75f07b22364258b8 

@martindurant suggested this could be solved by changing `consolidate` to a function rather than a module in https://github.com/fsspec/GSoC-kechunk-2022/issues/5 another option is to move postprocess to the beginning of the consolidate module. 

https://github.com/fsspec/kerchunk/blob/10908098cc5a179c580ae92bf0bd1c5d3165ffa8/kerchunk/combine.py#L382-L397

martindurant:
> could be solved by changing consolidate to a function rather than a module in https://github.com/fsspec/GSoC-kechunk-2022/issues/5 another option is to move postprocess to the beginning

These two are not mutually exclusive :)

PR #180, opened 2022-06-16
Run postprocess before consolidate in MZZ

peterm790:
This addresses issue https://github.com/fsspec/kerchunk/issues/176

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -65,7 +65,8 @@ class MultiZarrToZarr:
         Acts on the references dict of all inputs before processing. See ``drop()``
         for an example.
     :param postprocess: callable
-        Acts on output references dict before finally returning
+        Acts on the references dict before output. 
+        postprocess(dict)-> dict
     """
 
     def __init__(self, path, coo_map=None, concat_dims=None, coo_dtypes=None,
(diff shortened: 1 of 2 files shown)


martindurant:
I see you called nb strip on this repo. We have a conflict between wanting rendered notebooks so that they are more interesting to see when linked from the examples age, versus cutting down further PR sizes for any future pages. @rsignell-usgs , what do you think?

peterm790:
@martindurant or right I had setup nb-stripout to automatically run before committing, I will deactivate for the next commit.

Yes I am happy to have a look into the tests and why this failed

martindurant:
Now has different failures :)  Unfortunately, changing the internal API often has knock-on effects like this, where tests are written to poke those internal methods. At a glance, it seems simple enough to change the test itself to match the new code for one, but I don't know about test_postprocess, probably needs more investigation.

peterm790:
So this new commit passes all checks. I realised the initial way postprocess was written wasn't actually modifying `self.out`, which the test thankfully picked up on. Then I have made a small change to `test_combine` to call consolidate as a stand alone function rather than a module in `MultiZarrtoZarr`

kerchunk/combine.py
```diff
@@ -404,13 +388,32 @@ def translate(self, filename=None, storage_options=None):
             self.store_coords()
         if 3 not in self.done:
             self.second_pass()
-        out = self.consolidate()
+        if 4 not in self.done:
+            if self.postprocess is not None:
+                self.out = self.postprocess(self.out)
```
martindurant, line 393: Please alter the docstring for postprocess, to add its signature `postprocess(dict)->dict`.

PR #181, opened 2022-06-21
Set _FillValue

### LINKED ISSUE #177: FillValue issues with kerchunk 1e37=> 9.99999993e+36
In my original NetCDF4 file (HDF5 under the hood) the fill value `_FillValue` is set to `1.e+37` for 32bit floats, and when loaded into xarray with `decode_cf=True`, the data values with 1e37 are correctly interpreted as fill values and set to NaN. 

In the kerchunk JSON, the `_FillValue` in the JSON is set to 9.99999993e+36 and when loaded into xarray with `decode_cf=True`, the data values are not interpreted as fill values, and show up as 9.99999993e+36 (not 1e37 and not NaN). 

Pretty confused on what is going on here!

Here is a [Reproducible Notebook](https://nbviewer.org/gist/d66b81e9322b1216e496930a555f3cb2)

cc @peterm790

martindurant:
Fixes #177

### DIFF
--- a/kerchunk/hdf.py
+++ b/kerchunk/hdf.py
@@ -148,6 +148,9 @@ class SingleHdf5ToZarr:
             except TypeError:
                 lggr.debug(
                     f'TypeError transferring attr, skipping:\n {n}@{h5obj.name} = {v} ({type(v)})')
+        if isinstance(zobj, zarr.Array) and getattr(zobj, "fill_value") is not None:
+            lggr.debug(f"Udated _FillValue for {zobj}")
+            zobj.attrs["_FillValue"] = encode_fill_value(zobj.fill_value, zobj.dtype)
 
     def _translator(self, name: str, h5obj: Union[h5py.Dataset, h5py.Group]):
         """Produce Zarr metadata for all groups and datasets in the HDF5 file.
--- a/kerchunk/hdf.py
+++ b/kerchunk/hdf.py
@@ -132,7 +132,7 @@ class SingleHdf5ToZarr:
                 if v.dtype.kind == 'S':
                     v = v.astype(str)
                 if n == '_FillValue':
-                    v = encode_fill_value(v, v.dtype)
+                    continue  # strip it out!
                 elif v.size == 1:
                     v = v.flatten()[0]
                     if isinstance(v, (np.ndarray, np.number, np.bool_)):
@@ -148,9 +148,6 @@ class SingleHdf5ToZarr:
             except TypeError:
                 lggr.debug(
                     f'TypeError transferring attr, skipping:\n {n}@{h5obj.name} = {v} ({type(v)})')
-        if isinstance(zobj, zarr.Array) and getattr(zobj, "fill_value") is not None:
-            lggr.debug(f"Udated _FillValue for {zobj}")
-            zobj.attrs["_FillValue"] = encode_fill_value(zobj.fill_value, zobj.dtype)
 
     def _translator(self, name: str, h5obj: Union[h5py.Dataset, h5py.Group]):
         """Produce Zarr metadata for all groups and datasets in the HDF5 file.
@@ -208,6 +205,8 @@ class SingleHdf5ToZarr:
                 cinfo = self._storage_info(h5obj)
                 if h5py.h5ds.is_scale(h5obj.id) and not cinfo:
                     return
+                if h5obj.attrs.get("_FillValue") is not None:
+                    fill = encode_fill_value(h5obj.attrs.get("_FillValue"), dt or h5obj.dtype)
 
                 # Create a Zarr array equivalent to this HDF5 dataset...
                 za = self._zroot.create_dataset(


rsignell-usgs:
@martindurant , I thought we wanted to set `fill_value` to be the value of `_FillValue`, not vice versa:
https://github.com/fsspec/kerchunk/issues/177#issuecomment-1159098241

martindurant:
The code linked [in the issue](https://github.com/pydata/xarray/blob/d7931f9014a26e712ff5f30c4082cf0261f045d3/xarray/backends/zarr.py#L455-L456) is what I was going on

martindurant:
@rabernat to choose: when creating the zarr output metadata, fill_value overwrites _FillValue, or the reverse?

keewis:
as far as I understand https://github.com/fsspec/kerchunk/issues/177#issuecomment-1158882019, we should always ignore the HDF5 dataset fill value because it is static and not part of the netcdf4 spec.

keewis:
Well, https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html says:
> It is not necessary to define your own `_FillValue` attribute for a variable if the default fill value for the type of the variable is adequate. However, use of the default fill value for data type byte is not recommended.

So that means you'd use the default fill value defined in the netcdf4 library if `_FillValue` is not set?

For HDF5 datasets it seems the correct thing to do is what `kerchunk` currently does: translate the HDF5 `fill_value` property to `zarr`'s `fill_value`.

Re-reading the comments in #177, I get the impression that @ajelenak and @rabernat actually *disagree* on what to do (I might very well be misunderstanding something, though): @ajelenak would [prefer](https://github.com/fsspec/kerchunk/issues/177#issuecomment-1158882019) to expose both `fill_value` and `_FillValue` from `xarray`'s `zarr` backend and have `decode_cf` decide which to use, while @rabernat would [like](https://github.com/fsspec/kerchunk/issues/177#issuecomment-1158938067) to have `kerchunk` set `zarr`'s `fill_value` to `_FillValue` for netcdf4 datasets.

rabernat:
It's not a question of what I would like or not like. My goal was to communicate _how Xarray handles fill_value with Zarr_.

1. When opening a Zarr store for reading, the Zarr array `.fill_value` is copied into the CF-style attribute `_FillValue`. So when the data are decoded whatever the the `fill_value` is presented to the user as NaN. In [the code](https://github.com/pydata/xarray/blob/1d9c1e35a61cbb3680c141e29808c131e019301e/xarray/backends/zarr.py#L453-L456), you can see that any existing `_FillValue` attribute is overwritten. That is potentially a problem. However, that situation would never arise if you are using Xarray to write the data because...
2. When writing an array, the `_FillValue` attribute is _removed_ from attributes and placed in `fill_value` of the zarr array ([code](https://github.com/pydata/xarray/blob/1d9c1e35a61cbb3680c141e29808c131e019301e/xarray/backends/zarr.py#L606-L631)).

Creating a kerchunk Zarr is analogous to Xarray writing a Zarr store. So I recommend you follow the same pattern here. If `_FillValue` is present in the attributes, move it `.fill_value` of the array. Do not set the `_FillValue` attribute. Xarray will set it when decoding the data.

martindurant:
> detect the dataset type (netcdf or HDF)

I'm not sure this is a thing: all netCDF4 files are valid HDF5, but not vice versa; plus you can reasonably store netCDF datasets as *parts* of an HDF5 at some subpath.

keewis:
well, the point is that netcdf4 does not use / set the HDF5 fill value, as far as I can tell, so we need to figure out which fill value to use with a particular dataset. That "detection" does not have to happen in code, though, it could just as well be a parameter to `SingleHdfToZarr`.

rsignell-usgs:
@keewis I see your point -- if the HDF5 file is generated with NetCDF4, it will have `_FillValue` defined and we should use that for the Kerchunk `fill_value`,  but if not we should use the HDF5 `fill_value` for the Kerchunk `fill_value`.

martindurant:
How about this version, then? Would someone like to test?

rsignell-usgs:
Working!  https://nbviewer.org/gist/ae2dc327960c41e97a97e512bdc55a30

keewis:
> Is there a fast/definitive way to tell if an HDF5 file was generated by NetCDF4?

It is probably not definitive, but https://github.com/pedro-vicente/netcdf-detect uses some internal attributes that the netcdf library sets to detect that. It is written in C, but it should be possible to do the same using `h5py` (see also [this post](https://www.unidata.ucar.edu/mailing_lists/archives/netcdfgroup/2016/msg00092.html) and the follow-ups on the netcdf mailing list).

rsignell-usgs:
@keewis , so basically, check for:

1) an attribute named "_Netcdf4Dimid" (in some cases)
2) an attribute named "NAME", (always), saved by the HDF5 Dimension Scales API, that contains the string "This is a netCDF dimension but not a netCDF variable."

This utility tries to detect both attributes by traversing the HDF5 file, if either case is found, it returns a value of 1

PR #188, opened 2022-07-13
Set HDF5 API to closer match netCDF3

### LINKED ISSUE #186: API inconsistencies between SingleHdf5ToZarr and netcdf_recording_file
I am working on adding netCDF3 kerchunk support to Pangeo Forge. I noticed an inconsistency in the API which is making this difficult.

Nominally, these classes both do the same thing: take a netCDF file and translate it into a kerchunk reference file system. However,  `SingleHdf5ToZarr` has two required arguments:

> - h5f (file-like): Input HDF5 file as a binary Python file-like object (duck-typed, adhering to BinaryIO is optional)
> - url (string): URI of the HDF5 file.

..while `netcdf_recording_file` has only one

> - filename (str): location of the input

**Why?**

More specifically, why does SingleHdf5ToZarr require both a url AND a file-like object? Isn't that redundant? Conversely, if there is a reason to have these two separate arguments, why not support them both for `netcdf_recording_file`?

martindurant:
Fixes #186

@rabernat , this raises the question of what to do for the backends that are implemented as simple functions, e.g., grib. Should they all be wrapped in a class of the same structure?

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -2,7 +2,7 @@ name: Docs
 
 on:
   push:
-    branches:    
+    branches:
       - main
 
 jobs:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,5 +1,5 @@
 name: "Pull Request Docs Check"
-on: 
+on:
 - pull_request
 
 jobs:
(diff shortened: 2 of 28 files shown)


martindurant:
Merging this here, so that development can continue for netCDF3 without conflicts with the reformatting

PR #192, opened 2022-07-15
Add coordinate utility

martindurant:
@rsignell-usgs , a stab at geoTIFF coordinates, as a separate function not yet tied into kerchunking a TIFF. The intent is to produce coordinates as expected by xarray, values at the centre of each pixel (although RasterTypeGeoKey can specify whether we have pixels or points).
Care to compare with some geoTIFFs you have?


martindurant:
It's worth noting that TIFF reader doesn't necessarily produce netCDF output that xarray can read, even when xarray can read the original. For example, the lcmap_tiny_cog_2019.tif file used in the existing test produces a zarr group of three arrays for the multiple scales and no coordinates - apparently there is no similar logic to unbundle. loading the TIFF directly exposes only the high-res version. We should decide what's the correct representation for such data.

martindurant:
I added a [test](https://github.com/fsspec/kerchunk/pull/192/files#diff-8a4fd098c0d3e2c4a1396286aebd90dd0909074315c188a7b60c9821fd7543a0R32) and docstring to show what is supposed to happen here. Note that the output coordinate arrays are not stored anywhere, and a further extension might be to write them into an output reference JSON file.

martindurant:
I think this useful as it stands, though we will doubtless iterate on it yet. No hurry, though.

PR #193, opened 2022-07-15
add tutorial to docs and ERA5 example

peterm790:
I have moved the existing tutorial to 'Quick Start' and used the [extended ERA5 tutorial](https://github.com/peterm790/ERA5_Kerchunk_tutorial/blob/master/ERA5_tutorial_extended.ipynb) as the new tutorial.  I have also added a new ERA5 notebook to the examples as well as the associated reference json. Both these files are ~14MB.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -11,3 +11,4 @@ sphinx-rtd-theme
 nbsphinx
 nbconvert==6.0.0rc0
 ipywidgets
+pandoc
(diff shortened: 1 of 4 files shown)


martindurant:
@rsignell-usgs , what do you think: have the version in the documentation rendered, but the version linked from the examples be unrendered? Not sure what the standard is here.

martindurant:
ping @rsignell-usgs for a review

PR #196, opened 2022-07-18
Add merge_vars convenience function to combine

peterm790:
I have had a go at creating a convenience function to merge variables across datasets with identical coordinates. 

The initial code to open the files using `fsspec` is from `MultiZarrToZarr.fss` and seems to work fine although I have not provided a `target_options` argument as I wasn't sure how to do this. 

I am checking that the dimensions of the variables from the files are identical by comparing the shape of the array with the largest number of dimensions, which I assume to be the variable. This seems to work fine for the datasets I have tested against but it is maybe not the clearest. Another option is to have the user specify the `identical_dims` and have a check to ensure these are in fact identical.  Or is it okay to have no checks? 

Fundamentally all this is doing is merging the dictionary items across the jsons and so could have strange results if the dimensions are not identical.


kerchunk/combine.py
```diff
@@ -478,3 +478,21 @@ def _reorganise(coos):
     for k, arr in coos.items():
         out[k] = np.array(sorted(arr))
     return out
+
+def merge_vars(files):
+    #this merges variables across datasets with identical dimensions.
```
martindurant, line 483: Could use docs. Files is a list of dicts or list of str?

kerchunk/combine.py
```diff
@@ -478,3 +478,21 @@ def _reorganise(coos):
...
+        fo_list = []
+        for of in fsspec.open_files(files):
+            fo_list.append(of.open())
+    merged = ujson.load(fo_list[0])
+    z = zarr.open(merged['refs'])
+    shape = max([arr[1].shape for arr in z.arrays()], key=lambda t: len(t)) #returns shape of variable with most dims
+    for file in fo_list[1:]:
+        refs = ujson.load(file)['refs']
+        z = zarr.open(refs)
+        assert max([arr[1].shape for arr in z.arrays()], key=lambda t: len(t)) == shape, 'variable dimensions not identical'
```
martindurant, line 496: Probably just assume the user knows what they are doing? There might be a reason to have var1(x1, y1) and var2(x2, y2), i.e., completely independent from oneanother, in the same dataset.

martindurant (commented):
Could use a test :)

kerchunk/combine.py
```diff
@@ -478,3 +478,27 @@ def _reorganise(coos):
...
+    """Merge variables across datasets with identical coordinates
+
+    :param files: list(dict), list(str) or list(fsspec.OpenFile)
+        List of reference dictionaries or list of paths to reference json files to be merged
+    :param storage_options: dict
+        Dictionary containing kwargs to `fsspec.open_files`    
+    """
+    if isinstance(files[0], collections.abc.Mapping):
+        fo_list = files
+        merged = fo_list[0]
```
martindurant, line 492:
```suggestion
        merged = fo_list[0].copy()
```

kerchunk/combine.py
```diff
@@ -478,3 +478,27 @@ def _reorganise(coos):
...
+    """Merge variables across datasets with identical coordinates
+
+    :param files: list(dict), list(str) or list(fsspec.OpenFile)
+        List of reference dictionaries or list of paths to reference json files to be merged
+    :param storage_options: dict
+        Dictionary containing kwargs to `fsspec.open_files`    
+    """
+    if isinstance(files[0], collections.abc.Mapping):
+        fo_list = files
+        merged = fo_list[0].copy()
```
martindurant, line 492: I added `.copy()` here, so that the original input dictionary doesn't get changed. It is unlikely to be used again, but still...

kerchunk/tests/test_combine.py
```diff
@@ -548,3 +548,14 @@ def test_inline(refs):
...
+def test_merge_vars():
+    a = dict({"version":1,"refs":dict({"item1":1})})
+    b = dict({"version":1,"refs":dict({"item2":2})})
+    merge = kerchunk.combine.merge_vars([a,b])
+    assert list(merge['refs']) == ['item1', 'item2']
+    fs = fsspec.filesystem("memory")
+    fs.pipe("file1.json", b'''{"version": 1, "refs": {"item1": 1}}''')
+    fs.pipe("file2.json", b'''{"version": 1, "refs": {"item2": 2}}''')
+    merge = kerchunk.combine.merge_vars(['memory://file1.json', 'memory://file2.json'])
+    assert list(merge['refs']) == ['item1', 'item2']
```
martindurant, line 561:
```suggestion
    assert list(merge['refs']) == ['item1', 'item2']

```

martindurant:
I'll merge this when tests finish. We will eventually want to list this and any other similar utility functions in a dedicated API section of the docs.

Issue #194, opened 2022-07-16
Work around for CFGRIB memory leak?


emfdavid:
Has anyone found a work around for the memory leak in [CFGrib](https://github.com/ecmwf/cfgrib/issues/283)?

My best solution right now is running my scangrib calls in Dask and then get dask to kill the workers when they go over limit.

I am building a kerchunked version of HRRR output in GCP.
Anyone interested in access iff I can make it public?

martindurant:
I was not previously aware of the memory leak. grib_scan is due for an overhaul either way. One possibility if to use eccodes directory rather than cfgrib, as [gribscan](https://github.com/gribscan/gribscan) does. Likewise, [cogrib](https://github.com/TomAugspurger/cogrib/blob/main/cogrib.py) shows that it's possible to use cfgrib without local files, and presumably not write .idx files, if indeed that is the underlying issue.

Either/both of those approaches could be incorporated to what happens here, or you can use them as is to produce kerchunk-compatible output.

emfdavid:
Okay - if you figure out a path forward between those two and possibly fixing cfgrib -  I might be able to help implement it.
For now I will set low memory termination thresholds on my dask workers and add a periodic callback to delete the junk in /tmp. I filled up a 100GB VM in a few hours this afternoon.

martindurant:
Do you see the same leak when accessing the data after successful kerchunking? The codec `GRIBCodec` uses cfgrib on the file parts, so maybe it suffers the same problem.

In the long term, it is feasible to try to actually decode the contents of grib messages directly without cfgrib/eccodes, but the format is complex and that would take some effort.

PR #199, opened 2022-07-19
Cleanup temporary files

emfdavid:
Use context manager to cleanup temporary files.
Related to #194 
Still investigating other possible issues.
There are also temporary file issues in cfgrib.messages and eccodes.messages.

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -220,12 +220,13 @@ class GRIBCodec(numcodecs.abc.Codec):
 
     def decode(self, buf, out=None):
         buf = ensure_contiguous_ndarray(buf)
-        fn = tempfile.mktemp(suffix="grib2")
-        buf.tofile(fn)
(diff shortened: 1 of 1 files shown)


martindurant:
See also https://github.com/fsspec/kerchunk/pull/198 , although this is simple enough and useful to go in right away.

martindurant:
> There are also temporary file issues in cfgrib.messages and eccodes.messages

Do you mean, that when we make a message object from bytes, actually is makes a temporary file anyway? Ghastly!

emfdavid:
> Do you mean, that when we make a message object from bytes, actually is makes a temporary file anyway? Ghastly!

The temp files that cfgrib/eccodes makes are `foobar.idx` files... they do appear to be one to one with the messages... but I don't claim to understand how it works or what they contain. The code is mostly duplicated between cfgrib.messages and eccodes.messages. I am not sure why.

PR #208, opened 2022-08-05
Update tutorial

peterm790:
I have moved the existing tutorial to 'Quick Start' and used the [extended ERA5 tutorial](https://github.com/peterm790/ERA5_Kerchunk_tutorial/blob/master/ERA5_tutorial_extended.ipynb) to create a newer more thorough tutorial. 

This time created directly as a reStructuredText document to avoid pandoc. I will close https://github.com/fsspec/kerchunk/pull/193 if this builds successfully.


docs/source/tutorial.rst
```diff
@@ -0,0 +1,553 @@
...
+
+.. code:: 
+
+    import xarray as xr
+
+    %%time
+    fs_ = fsspec.filesystem("reference", fo='01_air_pressure_at_mean_sea_level.json', ref_storage_args={'skip_instance_cache':True},
+                           remote_protocol='s3', remote_options={'anon':True})
+    m = fs_.get_mapper("")
+    ds = xr.open_dataset(m, engine="zarr", backend_kwargs={'consolidated':False}, chunks={})
```
martindurant, line 79: You can do all this in one line, but its a bit messy:
```
ds = xr.open_dataset("reference://", engine="zarr", backend_kwargs={
    "consolidate": False,
    "storage_options": {"fo": '01_air_pressure_at_mean_sea_level.json', "remote_protocol": "s3",
        "remote_options": {"anon": True}}
)
```
(the complexity of this line is why people would eventually want to hide such a dataset prescription inside an Intake catalogue).

peterm790:
@martindurant I have used your above suggestion to open the datasets in a single line to try condense the code a little and hopefully make the tutorial easier to follow. I have also added a small section using an intake catalog. And included the example plots to give an idea of load times. 

I could go as far as defining a function to return a dataset at the beginning of the tutorial? I'm not sure how much this will condense things however. Something like:

      def return_ds(d):
          backend_args = {"consolidated": False, "storage_options": {"fo": d, "remote_protocol": "s3","remote_options": {"anon": True}}}
          ds =  xr.open_dataset("reference://", engine="zarr", backend_kwargs=backend_args)
          return ds

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+
+.. parsed-literal::
+
+    <xarray.Dataset>
+    Dimensions:                         (new_dimension: 2, time0: 744, lat: 721,
+                                         lon: 1440)
+    Coordinates:
+      * lat                             (lat) float32 90.0 89.75 ... -89.75 -90.0
+      * lon                             (lon) float32 nan 0.25 0.5 ... 359.5 359.8
+      * new_dimension                   (new_dimension) object '1' '2'
```
martindurant, line 233: and now this is the result of applying the regex to each input file name

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+      * time0                           (time0) datetime64[ns] 2020-01-01 ... 202...
+    Data variables:
+        air_pressure_at_mean_sea_level  (new_dimension, time0, lat, lon) float32 ...
+    Attributes:
+        institution:  ECMWF
+        source:       Reanalysis
+        title:        ERA5 forecasts
+
+
+Similarly we can map each file to a new variable using the special ``var`` character in coo_map. Here we use the same ``regex`` function but instead map these as new variables.
```
martindurant, line 243:
```suggestion
Similarly we can map each file to a new variable using the special ``var`` key in coo_map. Here we use the same ``regex`` function but instead map these as new variables.
```

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+    Data variables:
+        1        (time0, lat, lon) float32 ...
+        2        (time0, lat, lon) float32 ...
+    Attributes:
+        institution:  ECMWF
+        source:       Reanalysis
+        title:        ERA5 forecasts
+
+
+Another special character in ``coo_map`` is ``attr:``. This allows the user to access variables from the file attributes.
```
martindurant, line 278:
```suggestion
Another special character in ``coo_map`` is ``attr:``. This allows the user to access values from the each dataset's global attributes.
```

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+.. parsed-literal::
+
+    <xarray.Dataset>
+    Dimensions:  (time0: 1440, lat: 721, lon: 1440)
+    Coordinates:
+      * lat      (lat) float32 90.0 89.75 89.5 89.25 ... -89.25 -89.5 -89.75 -90.0
+      * lon      (lon) float32 nan 0.25 0.5 0.75 1.0 ... 359.0 359.2 359.5 359.8
+      * time0    (time0) datetime64[ns] 2020-01-01 ... 2020-02-29T23:00:00
+    Data variables:
+        ECMWF    (time0, lat, lon) float32 ...
```
martindurant, line 305: but both datasets had the save value of the attribute, so is this actually what we expect?

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+    Attributes:
+        institution:  ECMWF
+        source:       Reanalysis
+        title:        ERA5 forecasts
+
+
+The special character ``vattr:{var}:`` allows access to variable attributes. Here renaming the variable to instead use it’s short name.
```
martindurant, line 312:
```suggestion
The special value ``vattr:{var}:{attr}`` allows access to variable attributes. Here renaming the variable to instead use its short name. 
```

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+        ECMWF    (time0, lat, lon) float32 ...
+    Attributes:
+        institution:  ECMWF
+        source:       Reanalysis
+        title:        ERA5 forecasts
+
+
+The special character ``vattr:{var}:`` allows access to variable attributes. Here renaming the variable to instead use it’s short name. 
+
+There are a number of other special characters for ``coo_map`` documented in the `API reference <https://fsspec.github.io/kerchunk/reference.html#kerchunk.combine.MultiZarrToZarr>`__
```
martindurant, line 314: This line should follow the example block below

docs/source/tutorial.rst
```diff
@@ -0,0 +1,577 @@
...
+        gen_json(file)
+
+
+.. parsed-literal::
+
+    CPU times: user 30.4 s, sys: 4.74 s, total: 35.1 s
+    Wall time: 14min 44s
+
+
+The ``.json`` reference files we have generated can now be used to open virtual datasets through xarray. To achieve this it is first necessary to create a mapping of the reference file using ``fsspec``. Here specifying that the reference json is pointing to files stored on AWS.
```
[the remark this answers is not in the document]
peterm790: updated to now read:

The ``.json`` reference files we have generated can now be used to open virtual datasets through xarray or zarr. It is necessary to specify location of the reference ``json`` files, using the ``target_options`` argument, and the source data using the ``remote_options`` and ``remote_protocol`` arguments. Here specifying that the source data is stored on ``AWS S3`` and can be accessed anonymously.

docs/source/tutorial.rst
peterm790: In this case the `json_list` contains two files of consecutive time steps and the same variable

docs/source/tutorial.rst
```diff
@@ -23,7 +23,9 @@ Here we are considering Netcdf4 files and so use the kerchunk ``hdf`` module. Su
     from kerchunk.hdf import SingleHdf5ToZarr 
     import fsspec
 
-Using fsspec to create a pythonic filesystem, provides a convenient way to manage file urls.
+Using fsspec to create a pythonic filesystem, provides a convenient way to manage file urls. 
+
+The ``SingleHdf5ToZarr`` method takes both an ``h5f`` file and a ``url`` as input. The ``h5f`` file can either be a binary Python file-like object or a url, in which case it will be opened using ``fsspec`` and ``storage_options``. The ``url`` input is not used to open the file and is intended to allow the user to compute the reference files on data before it is uploaded to its final storage location. Thus the ``url`` input should be the url of the final file destination and not the current location.
```
peterm790, line 28: I have added this section after out conversation yesterday to make it clear the reference files do not need to be regenerated if the source files move. Although I am not sure how well this comes across here?

PR #210, opened 2022-08-17
Make sure scale/offset are numbers in netCDF3

martindurant:


martindurant:
Interestingly, these are applied by xarray, not zarr, which already loaded the data fine. That means I can't make the codec without *removing* the attributes so that xarray doesn't try to apply the scale again.

PR #217, opened 2022-09-01
Attempt simple concatenator

### LINKED ISSUE #216: concatenate zarr datasets
I would like to virtually concatenate a few zarr arrays along an axis. They are specifically not xarray datasets, and do not have coordinate arrays. 

Is this in scope for `kerchunk`s API?

martindurant:
Fixes #216


martindurant:
cc @ivirshup

martindurant:
Example:
```python
import fsspec
import numpy as np
import zarr

import kerchunk.combine
import kerchunk.zarr

x = np.arange(10)
x2 = np.arange(10, 20)
g = zarr.open("./out1.zarr")
g.create_dataset("x", data=x, chunks=(2,))
g = zarr.open("./out2.zarr")
g.create_dataset("x", data=x2, chunks=(2,))


ref1 = kerchunk.zarr.single_zarr("out1.zarr/", inline=0)
ref2 = kerchunk.zarr.single_zarr("out2.zarr/", inline=0)

out = kerchunk.combine.concatenate_arrays([ref1, ref2], path="x")

mapper = fsspec.get_mapper("reference://", fo=out)
g = zarr.open(mapper)
g.x[:]
```

ivirshup:
Awesome! This works fantastic, thanks!

I think some of the docs for the path argument got cut off.

kerchunk/combine.py
```diff
@@ -488,26 +488,78 @@ def _reorganise(coos):
...
+        if i == 0:
+            zdata = ujson.load(fs.open(f"{path}.zarray"))
+            shape = zdata["shape"]
+            chunks_per_file = int(shape[axis] / zdata["chunks"][axis])
+            shape[axis] *= len(files)
+            zdata["shape"] = shape
+            out[f"{path}.zarray"] = ujson.dumps(zdata)
+            for name in [".zgroup", ".zattrs", f"{path}.zattrs"]:
+                if name in fs.references:
+                    out[name] = fs.references[name]
```
ivirshup, line 556: Does this mean the result will always have the `.zattrs` of the last file?

martindurant: Actually of the first dataset (i == 0). They should all be the same.

ivirshup: Ah yeah, skipped that line 🤦

> They should all be the same.

Is this something that's normally checked?

martindurant: Could do, but I think a convenience function like this can leave it to the caller.

ivirshup: Yeah, I think it's fine for now, but it would be nice for it to be documented.

Maybe having a callback argument like `combine_metadata(key: str, entries: list[dict]) -> dict` could work, but also totally fine to leave simple for now.

martindurant: Yeah, I don't want to add more functionality, or else we creep towards another version of MultiZarrToZarr. The function as it is is useful, but also shows the kinds of things (along with merge_variables, just above) that can be done to directly manipulate references.

ivirshup (commented):
Found a couple issues with the shape and chunking:

* This assumes the shape of each array is the same
* This does not error if arrays have different chunking. I would prefer this errored rather than returning an invalid reference

martindurant:
I will add *optional* checks. I want to allow for the fast path where the user knows what they are doing. As part of the checks I could also look at the attributes, but I prefer not to, since it's not exactly clear what should be allowed or how to merge sets that differ.

Note that I expect there to be many of these situation-specific helper functions, ideally provided by users  :)

pbranson:
This is great, thanks for adding this feature Martin! I will try it out next time I have a use.

> * This does not error if arrays have different chunking. I would prefer this errored rather than returning an invalid reference

I agree also, there should not be a constraint on having heterogenous chunking within an array.

> I will add optional checks. I want to allow for the fast path where the user knows what they are doing. As part of the checks I could also look at the attributes, but I prefer not to, since it's not exactly clear what should be allowed or how to merge sets that differ.
> Note that I expect there to be many of these situation-specific helper functions, ideally provided by users :)

In this pattern I have found supressing expected warnings emitted during dask computations difficult to silence and can become painful. In a helper function, would using [logging](https://docs.python.org/3/howto/logging.html) be a better approach than raising a [warning](https://docs.python.org/3/library/warnings.html)?

ivirshup:
> I agree also, there should not be a constraint on having heterogenous chunking within an array.

I would agree, but `zarr` currently doesn't support this (though hopefully will in the near future). If this was supported, you would still need to say what the chunk sizes were correctly.

> these situation-specific helper functions

I guess I don't see this as being too situation specific. My impression of this library was that the fundamental operations were:

* Map from a chunked array store to a virtual file system
* Virtually concatenate arrays on the virtual  file system (`concatenate_arrays`)
* Merge virtual file systems (`merge_vars`)

Then there are high level helpers, like MultiZarrToZarr is for the netcdf/ xarray model. But that is a composition of the latter two operations and some logic for associating arrays with dimensions.

> I want to allow for the fast path

Are you thinking about this from the perspective of performance or low level control?

martindurant:
This is going in now.
If we find we need to point people to it, we can expose it in the API docs or mention it in tutorials. For now, it can be standalone as it is.

PR #222, opened 2022-09-26
low level grib

martindurant:
Totally scratch experimental code.

@TomAugspurger @d70-t , I thought you might be interested in this. The "complex packing" in grib doesn't look  too difficult to implement, but would probably need cython to work well. The only encoding I deal with here is jpeg2000.  I wonder if it's worth anyone's while to do this thoroughly, so that the end user won't need eccodes/cfgrib anymore (will still need them for getting the coords/attributes during scanning).

### DIFF
--- a/kerchunk/grib_low_level.py
+++ b/kerchunk/grib_low_level.py
@@ -1,0 +1,142 @@
+import io
+import numpy as np
+import fsspec
+
+import cfgrib
+import eccodes
+import numcodecs.abc
+import imagecodecs.numcodecs
+# imagecodecs.numcodecs.register_codecs()
+
+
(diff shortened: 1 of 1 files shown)


d70-t:
I think it would absolutely be great to have a standalone low level grib implementation :+1:. However, my fear might be, that this could become a very large project...
I've frequently encountered grib files, which use a mixture of GRIB1, GRIB2 and even undocumented extensions to GRIB1 which are only "specified" in the source code of eccodes. So supporting all (or even the most used subset) of GRIB can be a lengthy procedure.

Another issue might be, that it's quite common for those scale and offset numbers to change between every message. This essentially means that we would need per-chunk filters, if we want to combine multiple grib messages to one mutli-dimensional variable in a dataset.

PR #229, opened 2022-10-06
shonky fix for new cfgrib

### LINKED ISSUE #228: GRIB2 Test failing
As discovered in #227 - The tests for GRIB are failing on the CI.

I got the same issue once when trying to troubleshoot, but when I tried to run the following in the `kerchunk/tests/` dir, both statements come back as true:
```python
import numpy as np
import xarray as xr
from kerchunk.grib2 import scan_grib

fn = "./CMC_reg_DEPR_ISBL_10_ps10km_2022072000_P000.grib2"

out = scan_grib(fn)
ds = xr.open_dataset(
    "reference://",
    engine='zarr',
    backend_kwargs={
        'consolidated':False,
        'storage_options':{'fo':out[0]}
    }
)

ds2 = xr.open_dataset(fn, engine='cfgrib', backend_kwargs={'indexpath':''})

d1 = ds['latitude'].values
d2 = ds2['latitude'].values

print((np.isnan(d1) == np.isnan(d1)).all())

print((d1[~np.isnan(d1)] == d2[~np.isnan(d2)]).all())
```

martindurant:
Fixes #228

### DIFF
--- a/kerchunk/codecs.py
+++ b/kerchunk/codecs.py
@@ -96,6 +96,9 @@ class GRIBCodec(numcodecs.abc.Codec):
         finally:
             eccodes.codes_release(mid)
 
+        if var == "values" and eccodes.codes_get_string(mid, "missingValue"):
+            data[data == float(eccodes.codes_get_string(mid, "missingValue"))
+                 ] = np.nan
         if out is not None:
             return numcodecs.compat.ndarray_copy(data, out)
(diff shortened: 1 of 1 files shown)


lsterzinger:
Yeah this doesn't make much sense... I'll merge this fix into my PR to get the tests passing

Issue #235, opened 2022-10-18
Kerchunk and Zarr V3


jhamman:
The Zarr V3 spec is now undergoing public review and testing. This issue raises the question of how Kerchunk should integrate with the new spec.

Key changes in the V3 spec that are relevant particularly to Kerchunk (https://github.com/zarr-developers/zarr-specs/pull/149):
1. change to chunk and metadata key names
2. introduction of storage transformer extensions
3. likely introduction of a Sharding Storage Transformer extension: https://github.com/zarr-developers/zarr-specs/pull/152

Questions:
1. Has any work been done to produce kerchunk references that align with the v3 storage key conventions?
2. Could Kerchunk provide be thought of as a storage transformer in v3?
3. The Sharding proposal (linked above) includes some references to putting shards in hdf5 files. Could Kerchunk extend that spec?

martindurant:
Some of this I'll have to think about, but some things I can answer immediately.

- Kerchunk could produce v3 reference sets right now, and indeed convert v2<->v3 no problem, since it's only a rearrangement of paths. I don't think this would come with any benefit, though. No work has been done.
- I am not sure kerchunk can be a storage transformer rather than a storage provider. If yes, I don't see why it would be beneficial in itself. There would need to be more done in that transformer to be worth it.
- Yes, there is a thought to providing shards via kerchunk ( https://github.com/fsspec/kerchunk/issues/134 and preffs ) in a manner similar to but independent of the sharding spec.

I also want to mention that kerchunk should be useful for more than just zarr, so I will tend to favour things being coded in the storage layer rather than zarr-specific extensions. For example, reordering and selecting parquet files without touching the originals is something that kerchunk can do now. If you wanted full tabular iceberg compatibility using kerchunk/referenceFS, one could implement that now without too much trouble.

Here is the simplest non-zarr idea for CSVs: https://github.com/fsspec/kerchunk/issues/66 (and, more generally, random access of delimited/block compressed data).

martindurant:
? I thought one of the main reasons for having a V3 at all was so that we could have new things like sharding ?

That is presumably why my working variable-chunking implementation for v2 was not given consideration.

NikosAlexandris:
What is the status here ?

``` python
 /software/rekxx/rekx/parquet.py:75 in create_parquet_store
    72 │   │   log_messages.append(f"Created the filesystem mapper {output}")
    73 │   │
    74 │   │   log_messages.append(f"Kerchunking the file {input_file}")
[...]

PR #237, opened 2022-10-18
Zarr v3 support

jhamman:
This PR experiments with adding Zarr v3 support to Kerchunk. 

In most cases, this simply requires passing the `zarr_version` parameter to the `zarr.group` constructor. The primary exception is in `combine.py` where lots of v2 metadata keys are hard coded. I have not tried to tackle that part of the codebase yet.

If you are reviewing this, I suggest starting with the hdf5  translator and the Python 3.10 GitHub action.

towards #235


kerchunk/combine.py
```diff
@@ -81,6 +81,9 @@ class MultiZarrToZarr:
     :param postprocess: callable
         Acts on the references dict before output.
         postprocess(dict)-> dict
+    :param postprocess: int
```
rabernat, line 84:
```suggestion
    :param zarr_version: int
```

martindurant:
Thanks for kicking this off.

We will also need to do more work in the future, as v3 eventually gains more features that don't get backported to v2. For example, I don't know if we can already use the v3 unfinished sharding code to do what preffs does, concatenating multiple target byteranges to a single chunk for some of the data chunks. I suppose when v3 is finished and released, we could decide to drop v2 completely here, but I suspect it will be around for quite some time yet.

jhamman:
Somehow I messed up my fork here. I'm going to close this PR and I'll reopen.

Issue #248, opened 2022-10-28
Latitude and Longitude are not loaded in for GFS Grib2 datasets


mpiannucci:
I am using https://nbviewer.org/gist/peterm790/92eb1df3d58ba41d3411f8a840be2452 as a basis to kerchunk over NOAAs GFS Wave data on S3 here: https://noaa-gfs-bdp-pds.s3.amazonaws.com/index.html#gfs.20221028/06/wave/gridded/

Using that I created the following script to generate the reference jsons which works well: 

```python
import xarray as xr
import datetime as dt
import fsspec
import ujson
import hvplot.xarray
from kerchunk.grib2 import scan_grib
from kerchunk.combine import MultiZarrToZarr

fs_read = fsspec.filesystem('s3', anon=True, skip_instance_cache=True, use_ssl=False) # For now SSL false is solving my cert issues **shrug**
fs_write = fsspec.filesystem('')

hourly_files = fs_read.glob('s3://noaa-gfs-bdp-pds/gfs.20221028/06/wave/gridded/gfswave.t06z.atlocn.0p16.*.grib2')
files = sorted(['s3://'+f for f in hourly_files])
files

so = {"anon": True, "use_ssl": False}
json_dir = '../datasets/gfswave_atlantic/'

def make_json_name(file_url, message_number): #create a unique name for each reference file
    date = file_url.split('/')[3].split('.')[1]
    name = file_url.split('/')[7].split('.')[0:5]
    return f'{json_dir}{date}_{name[0]}_{name[1]}_{name[2]}_{name[4]}_message{message_number}.json'

def gen_json(file_url):
    print(f'Processing {file_url}...')
    out = scan_grib(file_url, storage_options=so)   # create the reference using scan_grib
    for i, message in enumerate(out): # scan_grib outputs a list containing one reference per grib message
        out_file_name = make_json_name(file_url, i)  # get name
        print(message)
        with fs_write.open(out_file_name, "w") as f: 
            f.write(ujson.dumps(message)) # write to file

for f in files[0:2]:
    gen_json(f)
```

However, when I load in one of the reference json files with xarray the coordinate arrays are just flat: 

```python
fs = fsspec.filesystem("reference", fo=reference_jsons[0], remote_protocol='s3', target_options={'anon': True}, remote_options={'anon': True, 'use_ssl': False})
mapper = fs.get_mapper('')

[...]

Issue #238, opened 2022-10-19
Problem using scan_grib with GEFS output


lucien-sim:
Thank you for all the work you've done on Kerchunk! The prospect of accessing data spread between tons of tiny model output files in a cloud-performant manner is really exciting! 

I'm currently running into some difficulties using Kerchunk to access GEFS output from AWS (https://registry.opendata.aws/noaa-gefs/). I'm following what I think is the "standard" pattern for accessing GRIB files (eg. https://nbviewer.org/gist/peterm790/92eb1df3d58ba41d3411f8a840be2452, https://github.com/fsspec/kerchunk/issues/150) but am getting a confusing error related to coordinates when I try to open the files with xarray. I've gathered that Kerchunk doesn't always play well with GEFS/GFS grib files and am happy to experiment with cogrib, but thought I'd put post the issue in case there's an easy fix. 

Here's my code: 

```python
import xarray as xr
import ujson
from kerchunk.grib2 import scan_grib
from kerchunk.combine import MultiZarrToZarr
import fsspec
import dask

# Set details for GEFS files, variables. 
# For simplicity, only want first 6 hours for a single variable and ensemble
# member (control) in a single run. 
run_date = '20221011'
cycle = '00'
vbl_collection = 'pgrb2ap5'
type_of_level = 'heightAboveGround'
level = 2
grib_name = 't2m'
fhr_range = (0, 6)
json_dir = '/Users/lucien/Documents/projects/kerchunk'

s3_so = {
    'anon': True, 
    'skip_instance_cache': True
}

afilter = {
    'typeOfLevel': type_of_level, 
    'level': level, 
    'cfVarName': grib_name
}

s3_fs = fsspec.filesystem('s3', **s3_so)
local_fs = fsspec.filesystem('file', auto_mkdir=True)

# Get list of remote files. 
files = [
    f"s3://{f}"
    for f in s3_fs.glob(f"s3://noaa-gefs-pds/gefs.{run_date}/{cycle}/atmos/{vbl_collection}/*")
    if f[-4:] != '.idx'
    and '/gec00.' in f
    and fhr_range[0] <= int(f[-3:]) <= fhr_range[1]
]

print(files)

def make_json_name(file_url: str, message_number: int) -> str: 
    grib_name = file_url.split('/')[-1]
    return f"{json_dir}/{grib_name}_message{message_number}.json"


def gen_json(file_url: str) -> None:
[...]

PR #249, opened 2022-10-28
Fix gridding for regular_ll grib2 grid types

mpiannucci:
See https://github.com/fsspec/kerchunk/issues/248 for discussion. 

`regular_ll` grid types in grib2 messages are (lat,lng) indexed

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -175,10 +175,13 @@ def scan_grib(
                 _store_array(
                     store, z, data, name, inline_threshold, offset, size, attrs
                 )
+
             dims = (
                 ["x", "y"]
                 if m["gridType"] in cfgrib.dataset.GRID_TYPES_2D_NON_DIMENSION_COORDS
-                else ["longitude", "latitude"]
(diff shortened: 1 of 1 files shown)


martindurant:
So... are there grid types where it should be the original order, which seemed to work elsewhere - is it just regular_ll that's odd?

mpiannucci:
Yeah thats my question too. The only grib models I work with are all `regular_ll`. HRR is lambert, do you know any other regular gridded grib files that worked previously?

martindurant:
Prsumably @lucien-sim 's in #238

mpiannucci:
Okay modifying my original script to read in the gefs instead allows me to read and plot without issues: 

Screen Shot 2022-10-28 at 3 17 10 PM

Screen Shot 2022-10-28 at 3 17 15 PM

Would be interested if it looks different than what he was seeing though

martindurant:
That definitely looks like ocean

rsignell-usgs:
@mpiannucci I'm so excited you are contributing here!

PR #255, opened 2022-11-23
Start auto dasker combine

martindurant:

PR #256, opened 2022-11-28
Update hdf.py

peterm790:
I usually open the file first before passing to SingleHdf5ToZarr so haven't encountered this before, this change was needed to open a file from s3 from within SingleHdf5ToZarr. I assume this is a bug and not done this way on purpose?

### DIFF
--- a/kerchunk/hdf.py
+++ b/kerchunk/hdf.py
@@ -89,7 +89,7 @@ class SingleHdf5ToZarr:
         if vlen_encode not in ["embed", "null", "leave", "encode"]:
             raise NotImplementedError
         self.vlen = vlen_encode
-        self._h5f = h5py.File(h5f, mode="r")
+        self._h5f = h5py.File(self.input_file, mode="r")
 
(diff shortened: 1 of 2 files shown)


martindurant:
Can we please make a test, where the file is held at a "memory://path.hdf" URL?

martindurant:
In #255 I updated the CI to use mamba instead of conda. Maybe I merge that now and defer further work on docs for auto-dask, so you can merge it into here and get CI to pass.

peterm790:
I have added a test, which just ensures a file opened directly within SingleHdf5ToZarr is the same as one where an OpenFIle is passed. 
I assume the linting is failing because I have recycled some code from another test? Should I combine the `test_single` and `test_single_direct_open` into a single test?

martindurant:
It was simply formatting. I recommend running `pre-commit install` in the repo directory to have those things run each time before you push. There are also github actions to do the fixes automatically, but I haven't set those up.

PR #258, opened 2022-12-01
Contribution Guide

norlandrhagen:


martindurant:
ping me when you want a review

norlandrhagen:
@martindurant, whenever you get a chance, a review would be great!

martindurant:
I would take pains in a document like this to point out that kerchunk is not an end-user library, but a power-user tool (end users get the artefacts and may never install kerchunk). Also, that it is far from mature, and it's OK to make major breaking changes.

norlandrhagen:
@martindurant I incorporated your suggestions. Any other thoughts before this is merged?

docs/source/contributing.rst
```diff
@@ -0,0 +1,242 @@
...
+.. note::
+
+  Large parts of this document are based on the
+  `Xarray Contributing Guide <http://docs.xarray.dev/en/stable/contributing.html>`_
+  , the `Pandas Contributing Guide <http://pandas.pydata.org/pandas-docs/stable/contributing.html>`_
+  and the `xbatcher Contributing Guide <https://xbatcher.readthedocs.io/en/latest/contributing.html>`_.
+
+.. warning::
+    kerchunk is actively in development and breaking changes may be introduced at any point.
+    In its current state, it is possible for experienced users to build kerchunk indices of datasets, where end users never have to touch kerchunk.
```
martindurant, line 16:
```suggestion
    In its current state, it is possible for experienced users to build kerchunk indices of datasets,
    which work today and will continue to work. End users of these reference sets will not be exposed
    to changes in kerchunk's code and usually don't even need to install kerchunk. 
```

Issue #259, opened 2022-12-06
Grib2 scan_grib: Dimensions are Out of Order


keltonhalbert:
I've been attempting to follow [this kerchunk grib2 example using HRRR data](https://nbviewer.org/gist/peterm790/92eb1df3d58ba41d3411f8a840be2452), but I have encountered an interesting issue that prevents me from being able to use the Dataset in my intended manner due to the mis-naming (or mis-ordering) of X and Y dimensions for arrays. 

Running the example_combine from the documentation (copied below) and printing the dataset shows the issue: the arrays are labeled as ```(time, x, y)```, but the actual ordering of the grib2 data is ```(time, y, x)```. 1799 is the X extent of the domain and 1059 is the Y extent of the domain, yet in the dimensions, they are labeled opposite.  Attempting to perform operations on the Dataset when working with X and Y coordinate names doesn't work because of this behavior. 

I've attempted manually modifying the kerchunk/grib2.py file to re-order occurrences of ```["x", "y"]```, but sadly the solution doesn't appear to be that simple. Any ideas on how to make sure that the array dimensions are labeled correctly? 

## Script Output
```
<xarray.Dataset>
Dimensions:            (time: 1, x: 1059, y: 1799, heightAboveGround: 1)
Coordinates:
  * heightAboveGround  (heightAboveGround) int64 2
  * time               (time) datetime64[ns] 2019-01-01T22:00:00
Dimensions without coordinates: x, y
Data variables:
    2d                 (time, x, y) float64 dask.array<chunksize=(1, 1059, 1799), meta=np.ndarray>
    2r                 (time, x, y) float64 dask.array<chunksize=(1, 1059, 1799), meta=np.ndarray>
    2sh                (time, x, y) float64 dask.array<chunksize=(1, 1059, 1799), meta=np.ndarray>
    2t                 (time, x, y) float64 dask.array<chunksize=(1, 1059, 1799), meta=np.ndarray>
    latitude           (x, y) float64 dask.array<chunksize=(1059, 1799), meta=np.ndarray>
    longitude          (x, y) float64 dask.array<chunksize=(1059, 1799), meta=np.ndarray>
    pt                 (time, x, y) float64 dask.array<chunksize=(1, 1059, 1799), meta=np.ndarray>
Attributes:
    centre:             kwbc
    centreDescription:  US National Weather Service - NCEP
    edition:            2
    subCentre:          0
```

## Example Script
```
import xarray as xr
import fsspec
[...]

PR #267, opened 2022-12-13
Fix for #248, #249, and #259, along with variable naming change

keltonhalbert:
### About the Change
See #259 for discussion, and #249 for history of issue/changes being reverted or modified. 

In addition to the issues with the dimension labels being wrong, this PR includes a change to variable names passed to xarray. Xarray and cfgrib tend to use the ```cfVarName``` if its available, and then falls back to the grib ```shortName``` if its unavailable or unknown. This change should mean that xarray variable names are consistent across kerchunk/cfgrib/xarray.

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -161,8 +161,8 @@ def scan_grib(
                 + cfgrib.dataset.EXTRA_DATA_ATTRIBUTES_KEYS
                 if k in m
             }
-            ## try to use cfVarName if available,
-            ## otherwise use the grib shortName
+            # try to use cfVarName if available,
+            # otherwise use the grib shortName
             varName = m["cfVarName"]
             if varName in ("undef", "unknown"):
                 varName = m["shortName"]
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -177,11 +177,9 @@ def scan_grib(
                     store, z, data, name, inline_threshold, offset, size, attrs
                 )
             dims = (
-                ["x", "y"]
+                ["y", "x"]
                 if m["gridType"] in cfgrib.dataset.GRID_TYPES_2D_NON_DIMENSION_COORDS
                 else ["latitude", "longitude"]
-                if m["gridType"] in cfgrib.dataset.GRID_TYPES_DIMENSION_COORDS
-                else ["longitude", "latitude"]
             )
             z[m["shortName"]].attrs["_ARRAY_DIMENSIONS"] = dims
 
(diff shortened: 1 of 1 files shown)


martindurant:
So if this explicitly fixes the projections where lat and lon separate, do we also have a test case for where they are 2D lookup tables?

keltonhalbert:
> So if this explicitly fixes the projections where lat and lon separate, do we also have a test case for where they are 2D lookup tables?

I think this should. If my understanding isn't too terribly flawed, the HRRR grib2 files with the Lambert grid use 2D lookup tables, as latitude and longitude cannot be given as 1D array axes.

martindurant:
> Edit: Would it be helpful to include a test pulling from the HRRR data on S3?

In principle yes, and we do this for some HDF5 file. Of course, better if the target is small and so less likely to fail due to network timeouts.

keltonhalbert:
> > Edit: Would it be helpful to include a test pulling from the HRRR data on S3?
> 
> In principle yes, and we do this for some HDF5 file. Of course, better if the target is small and so less likely to fail due to network timeouts.

Awesome, I can work on implementing some tests for grib2. Looking at the grib test file, it looks like there isn't a whole lot going on. What are the priority checks you guys would like to see? 
https://github.com/fsspec/kerchunk/blob/f4159d0c5ed8f504aa7fb2972c4377afbaefbf10/kerchunk/tests/test_grib.py

For the 2D lat/lon grids (i.e. Lambert like HRRR), we can make sure their shapes are equal to the shapes of the other fields... but there are no 1D  coordinates to check the lat/lon shapes against for non-LL data unless you look at the cfgrib Nx and Ny fields. Alternatively, with xarray/cfgrib, when parsing a grib file cfgrib will automatically add a coordinate X and Y dimension that is 1D, with values from 0 - Nx-1 (or Ny-1). Should we do the same within scan_grib in order to ensure the sanity of the dimensions? Or should the test be as simple as loading a file with scan_grib, and then loading a file with xarray/cfgrib, and then comparing the dimensions, dimension lengths, and variable names?

As far as testing data, a standard HRRR grib file is ~140 MB. The tests could use scan_grib to load just a subset (one or two fields) from the dataset to do the checks on. Then perhaps do the same thing with a GFS file, and call it a day?

martindurant:
@keltonhalbert , do you think you can add a test as I suggest? I would contribute to this PR, but it is coming from your main branch, so I won't push there.

martindurant:
I added tests in a new PR rather than step on this one. Please review #276

PR #269, opened 2022-12-14
update to docs reqs

norlandrhagen:
Should fix dependency issues caused by this PR: https://github.com/fsspec/kerchunk/pull/258

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx-rtd-theme
+pangeo-sphinx-book-theme
 numpydoc==1.2.1
 autodoc
 numpy


norlandrhagen:
Ah yup, the theme would change the look of the docs a bit. It is the pangeo-sphinx-book-theme, which pangeo-forge and a few other project use (https://github.com/pangeo-forge/pangeo-sphinx-book-theme). 

Happy to pull it out into a separate PR and move the chat there if that is helpful.

martindurant:
This being an fsspec library, I would prefer to maintain the same look as the other fsspec libraries, RTD basic.

norlandrhagen:
Sounds good, I'll revert it.

martindurant:
This PR actually broke the CI build...

maxrjones:
> This PR actually broke the CI build...

as you're probably aware, https://github.com/fsspec/kerchunk/pull/271 fixed the CI issue caused by the sphinx html_theme setting.

@norlandrhagen, https://github.com/fsspec/kerchunk/blob/main/ci/docs.yml will also need changes. Apologies for not bringing this up before #258 was merged, but my opinion is that it would actually be better to remove that file entirely because it is redundant with https://github.com/fsspec/kerchunk/blob/main/docs/requirements.txt, which is required for the [GitHub action that's used to build the kerchunk docs](https://github.com/ammaraskar/sphinx-action). If you submit a PR to remove that file, you could also update the instructions in the contributing guide to use the docs/requirements.txt file:

```
conda create --name kerchunk-docs python=3.8
conda activate kerchunk-docs
python -m pip install -r docs/requirements.txt
```

norlandrhagen:
Ah thanks for catching that @maxrjones! I'll open up a PR.

Issue #209, opened 2022-08-08
Upcoming major improvements


martindurant:
Stuff that would be cool to get done and is well within out capacity. Please vote if you have favourites here!

- [x] automagically run combine in dask tree reduction mode #255
- [x] split to subchunk utility for uncompressed data. #251
- [ ] concatenate from subchunks for irregular inputs and primitive sharding support
- [x] coords generation (e.g., FITS WCS, geoTIFF bounding boxes) for flexible xarray indexes (partial: #192 )
- [ ] parquet storage (from [preffs](https://github.com/d70-t/preffs)) #277 with 
  - [ ] lazy loading
  - [ ] sorted key partitioning 
- [ ] subsampling and multiscale pyramid generation
- [ ] checksums or other UUIDs for remote files, to check for changes
- [ ] linked zarr/numcodecs: allow for read context to be passed to codec and storage layer, so that we can, for instance, apply a different scale to a cube at each pane https://github.com/zarr-developers/zarr-python/pull/1131
- [x] consolidate (nearly) adjoining reads in ReferenceFileSystem as is done in fsspec.parquet https://github.com/fsspec/filesystem_spec/pull/1063

nabobalis:
I would vote for "coords generation" especially with FITS.

rsignell-usgs:
1. "coords generation" for GeoTIFF
2. handling netcdf files with different `scale_factor` and `add_offset`
3. parquet

emfdavid:
`automagically run combine in dask tree reduction mode` for me please!

joshmoore:
"parquet storage (from [preffs](https://github.com/d70-t/preffs))" this sounds nifty, but I'll add for what it's worth, I did discuss with @jakirkham what it would look like to use zarr for the storage of kerchunk itself :smile:

martindurant:
I would certainly love your ideas, and the thought had certainly occurred to me. 

In favour of parquet:
- the data is essentially tabular, no benefit from chunking on higher dimensions
- must of data data (key and embedded chunks or metadata) are str/bytes. OTOH, the keys (required for every entry) could be fixed-string
- partitioning into unequal-sized pieces, allowing, for instance, all of the references of one variable to live together and only be loaded at need; the column min/max values in the metadata also help with this
- in preffs, each key may appear multiple times, to indicate concatenation of subchunks. Parquet could maybe also achieve this with variable-length lists of references. I'm unconvinced that this is a good idea, but zarr doesn't have the capability.

[...]

PR #272, opened 2022-12-15
combine with low memory

martindurant:
This loads all the filesystems to be combined *twice*. If successful, may provide a flag for high versus low memory operation.

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -1,4 +1,5 @@
 import collections.abc
+import gc
 import logging
 import re
 from typing import List
@@ -252,6 +253,7 @@ class MultiZarrToZarr:
                     coos[var].update(value)
                 else:
                     coos[var].add(value)
+            del z, fs
 
         self.coos = _reorganise(coos)
         for c, v in self.coos.items():
@@ -262,6 +264,7 @@ class MultiZarrToZarr:
                 )
         logger.debug("Created coordinates map")
         self.done.add(1)
+        gc.collect()
         return coos
 
     def store_coords(self):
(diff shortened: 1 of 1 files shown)


rsignell-usgs:
I tried using this PR but it seems to take the same amount of memory and be 20% or so slower

martindurant:
@rsignell-usgs , maybe better now?

rsignell-usgs:
@martindurant , yes, better!   You [reported that combining 13 years took 13GB max of memory](https://github.com/fsspec/kerchunk/issues/209#issuecomment-1353647763) with the version in this PR:  

Here's what I got, combining 13 years:
* using this PR, version `0.0.1+420.gca577c4.dirty':  19GB max, Wall Clock 13min45s
* using the latest version from Main `0.0.9+69.g9f9076c.dirty`:  crashed server w/ 32GB memory

So better memory usage, but still requires more than many JupyterHub users will have.   

And I still can't create a single combined JSON for my dataset on the ESIP Qhub, even with the low-memory PR.   I want to combine the 40 years, but I can't even combine 20 without blowing out the 32GB memory!

martindurant:
Can you do 4x10 years, though, and then combine those?

rsignell-usgs:
@martindurant , no, I can't.  I can create the four decadal JSONs, but I can't combine even 2 of them with 32GB memory:
https://nbviewer.org/gist/cce070d4917aeb8d837c7259e3e6ff1a

martindurant:
@rsignell-usgs , I don't know if you can see my qhub "Untitled.ipynb". We can talk about this sometime.

rsignell-usgs:
With the new parquet lazy loading approach, I can load this dataset comfortably on an 8GB memory JupyterHub.   Hurrah!

Here's the proof:
https://nbviewer.org/gist/rsignell-usgs/77609e86825c9c6163af15f90590b440

martindurant:
Yep, so the combine step was "standard" and needed loads of RAM, and the resulting dict/JSON output was then transformed to parquet. We do not have a way to combine directly with parquet right now, but at least the final output was usable.

Issue #34, opened 2021-06-24
fletcher32 checksum


tedhabermann:
All datasets with fletcher32 == True are rejected by referenceMaker. This is the checksum supported by HDF5. 
Would it be possible to add it to nomcodecs? @ajelenak

ajelenak:
I think so. numcodecs already supports Adler32 which is similar to Flether32, I think.

martindurant:
The [code](https://github.com/UCSantaCruzComputationalGenomicsLab/hdf5/blob/master/src/H5checksum.c#L112) looks simple enough. 

The question is, how is this value included in the target bytes? Is it just in metadata around the bytes? If yes, then we can already ignore the checksum and use the bytes; but implementing the function above would enable the (optional) validation.

martindurant:
Example code of how to fetch the checksums: https://forum.hdfgroup.org/t/accessing-fletcher32-checksum/7299/3

PR #274, opened 2022-12-19
Parametrized tests for netcdf encoding options

rabernat:
This PR adds a test to verify that kerchunk can decode all possible flavors of netcdf4.

At least in my local dev environment, this reveals that fletcher32 decoding is broken, despite #34.

xref https://github.com/pydata/xarray/issues/7388

### DIFF
--- a/kerchunk/tests/test_hdf.py
+++ b/kerchunk/tests/test_hdf.py
@@ -286,3 +286,40 @@ def test_compact():
     m = fsspec.get_mapper("reference://", fo=out)
     g = zarr.open(m)
     assert np.allclose(g.ancillary_data.atlas_sdp_gps_epoch[:], 1.19880002e09)
+
+
+@pytest.mark.parametrize("zlib", [True, False], ids=["zlib", "no_zlib"])
+@pytest.mark.parametrize("shuffle", [True, False], ids=["shuffle", "no_shuffle"])
+@pytest.mark.parametrize("fletcher32", [True, False], ids=["fletcher32", "no_fletcher32"])
+def test_encoding_options(zlib, shuffle, fletcher32, tmp_path):
+    fname = tmp_path / "test.nc"
+    
+    shape = (2, 10)
+    chunksizes = (1, 10)
+
+    encoding = {
+        'zlib': zlib,
+        'shuffle': shuffle,
+        'complevel': 2,
+        'fletcher32': fletcher32,
+        'contiguous': False,
+        'chunksizes': chunksizes
+    }
+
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks for making sure this got tested.
The documentation misled me: the 4-byte checksum is added to the byte buffer before compression, not after, it seems.

rabernat:
Awesome! Great to hear that you can see a solution. Feel free to push to my branch if you want to just fix it in this PR.

martindurant:
Already did :)

rabernat:
😆  you're too fast Martin!

A great next step would be to actually implement the real checksum filter in numcodecs, rather than a dummy passthrough.

martindurant:
Agree - but kerchunk is all about fast and functional!

https://en.wikipedia.org/wiki/Fletcher%27s_checksum#Fletcher-32

https://gist.github.com/AJPoulter-Soton/9d0d2505af64f0719bdee59b9a4533ba ?

rabernat:
I can't imagine that computational cost of fletcher32 would be comparable to the actual decompression step (not to mention the i/o itself).

kerchunk/codecs.py
```diff
@@ -66,6 +66,19 @@ def decode(self, buf, out=None):
...
 
 
+class FletcherDummyFilter(numcodecs.abc.Codec):
+    codec_id = "fletcher_null"
+
+    def decode(self, buff, out=None):
+        return buff[:-4]
+
+    def encode(self, buf):
+        pass
```
rabernat, line 76: Wouldn't you want to append 4 empty bytes here, just in case someone tried to use it in a round-trip context?

martindurant: Hm, I'm not sure. Maybe raising NotImplemented is the correct thing to do, as the output can't be used by anything that really does the check.

martindurant:
Incoming new codec: https://github.com/zarr-developers/numcodecs/pull/412 (so there would be no need for one here once that is in)

martindurant:
I can change this to point to the new fletcher codec in numcodecs. Is the null version, skipping the CPU time to do the comparison, useful? I suppose fletcher ought to be pretty cheap to compute.

rabernat:
I personally consider it quite dangerous to pass through the data without actually checking the checksum. Since we now have a fast fletcher32 codec in numcodecs, I see no reason to keep the passthrough codec.

Profiling would be easy to do if we want some actual numbers.

martindurant:
We'll need a numcodecs release anyway

PR #283, opened 2023-01-24
Do not use auto HDF5 fill value for netcdf-derived variables

martindurant:


martindurant:
I'm OK to merge it, but if there are any case studies likely to be tried in the next few days, I would wait.

rsignell-usgs:
@pnorton-usgs, are there any other tests you might have handy?

pnorton-usgs:
I re-kerchunked our National Hydrologic Model output and verified that the HDF5 fill value is no longer used in the output json. The datatypes that were being converted to float are now remaining integers. Thanks for the quick fix on this one!

PR #287, opened 2023-01-26
Add docstrings and API ref

martindurant:


martindurant:
@rsignell-usgs , if you have a chance to review.

PR #290, opened 2023-01-31
mod doc on time coord generation from filenames

rsignell-usgs:
Was just going to correct a typo, but ended up cleaning up the language on generating time coordinates from filename info.


pl-marasco:
@rsignell-usgs Thanks, as I'm not a native speaker I appreciate your help on this. Now seems to be more clear.

PR #292, opened 2023-02-02
Zarr v3 support

jhamman:
This PR experiments with adding Zarr v3 support to Kerchunk.

In most cases, this simply requires passing the zarr_version parameter to the zarr.group constructor. I've decided to leave multizarr out of this PR for now. Its going to take more work and I'd like to avoid the larger refactor that may be needed for now.

If you are reviewing this, I suggest starting with the hdf5 translator and the Python 3.10 GitHub action.

towards https://github.com/fsspec/kerchunk/issues/235
second try after https://github.com/fsspec/kerchunk/pull/237


jhamman:
> Am I right to think that V3 reference sets will currently fail in MultiZarrToZarr? Do you have an estimate of how complex it would be to support?

Yes, I started down the path of reworking this but I found it to be too complicated. There are just so many places to special case key names that I wasn't sure it was worth the effort.

martindurant:
> Yes, I started down the path of reworking this but I found it to be too complicated. There are just so many places to special case key names that I wasn't sure it was worth the effort.

I wonder if it might be simpler to create a v2->v3 converter. Of course, in the long run we want v3 to become the norm.

jhamman:
Closing this as a) super out of date and b) not in line with the v3 spec. I would encourage those following this PR to revisit the v2/v3 conversation after zarr-python 3 is released.

Issue #293, opened 2023-02-03
Using Parquet for large references 


rsignell-usgs:
I've been testing out DFReferenceFileSystem, which loads the references from a collection of Parquet files (one Parquet file for metadata and one for each variable) instead of from a single JSON file.

Two questions:
1. I understand the references are only loaded when a specific variable is requested, but then all the references for that variable are loaded, correct?    (Or does it load just the references for the subset requested)?
2. I tried loading some data from a specific time step into a dataarray, and then replaced that dataarray with data loaded from another variable at the same time step.   Each time I switched the variable, I saw the memory usage increasing, and after changing the variable four times, I blew out my 8GB memory.   Is it possible to purge the references from memory when you no longer using them?

martindurant:
Correct, atm., all the references for a variable are loaded on first access, and kept around. 

A purgeable cache would be very reasonable, as would be splitting each variable into reference chunks to allow even more partial loading. These are the principal benefits of @agoodm 's approach, which could/should be adopted.

agoodm:
@rsignell-usgs since we are on the subject, you can give my approach a try with your test dataset. The code is currently in this [gist](https://gist.github.com/agoodm/25d41ce0c47cd714271be66d0db0459d) at the moment and you can use it as follows (making sure you have the latest development version of fastparquet installed as well):

```python
import fsspec
import xarray as xr

# refs is original references JSON dict
make_parquet_store('refs_test', refs)
mapper = ParquetReferenceMapper('refs_test')
fs = fsspec.filesystem('reference', fo=mapper, remote_protocol='file')
ds = xr.open_zarr(fs.get_mapper(''))
```
This uses an LRU cache along with splitting the references into row groups (of 1000 references by default) to help keep the in-memory footprint down. More details are provided [here](https://discourse.pangeo.io/t/trick-for-improving-kerchunk-performance-for-large-numbers-of-chunks-files/3090).

martindurant:
Thanks for the snippet! Also interested to see how well it fares.

Also interesting would be to see whether we can convert the individual files' JSON data too, and see an improvement in the memory footprint of the combine step - that would be a big win too.

rsignell-usgs:
Progress report:  I created a large enough instance on our AWS JupyterHub to run the code, but I can't build the latest fastparquet from github/main because the JupyterHub doesn't have gcc.   @martindurant says he can cut a dev release later today so that the build on conda-forge will trigger.

rsignell-usgs:
[...]

PR #298, opened 2023-02-16
Lazy Parquet References

agoodm:
This is a follow-up to the discussion in #293 and this [pangeo discourse thread](https://discourse.pangeo.io/t/trick-for-improving-kerchunk-performance-for-large-numbers-of-chunks-files/3090) to better address the current performance issues with having very large numbers of references. A quick overview of the changes: For now I have left the previous version of `refs_to_dataframe` along-side my new changes in `make_parquet_store` for reference, but presumably the end result of this PR is merging these two into just the former.

A quick todo list:
- [x] Support separate dtype columns
- [ ] preffs style support (multiple paths per chunk key)?
- [x] Store URLs categorically

### DIFF
--- a/kerchunk/df.py
+++ b/kerchunk/df.py
@@ -4,6 +4,7 @@ import ujson
 import os
 import pandas as pd
 import fsspec
+import zarr
 
 from kerchunk.utils import templateize
 
@@ -28,13 +29,15 @@ def _proc_raw(r):
     return r
 
 
-def get_variables(keys):
+def get_variables(refs, consolidated=True):
     """Get list of variable names from references.
 
     Parameters
     ----------
-    keys : list of str
+    refs : dict
         kerchunk references keys
+    consolidated : bool
+        Whether or not to add consolidated metadata key to references. (default True)
 
     Returns
     -------
@@ -42,7 +45,10 @@ def get_variables(keys):
         List of variable names.
     """
     fields = []
-    for k in keys:
+    meta = {}
+    for k in refs:
+        if ".z" in k and consolidated:
+            meta[k] = refs[k]
         if "/" in k:
             name, chunk = k.split("/")
             if name not in fields:
@@ -51,6 +57,10 @@ def get_variables(keys):
                 continue
         else:
             fields.append(k)
+    if consolidated and ".zmetadata" not in fields:
+        zarr.consolidate_metadata(meta)
+        refs[".zmetadata"] = meta[".zmetadata"]
+        fields.append(".zmetadata")
     return fields
 
 
@@ -87,6 +97,7 @@ def _write_json(fname, json_obj):
 def refs_to_dataframe(
     refs,
     url,
+    consolidated=True,
     storage_options=None,
     row_group_size=1000,
     compression="zstd",
@@ -106,6 +117,8 @@ def refs_to_dataframe(
     url: str
         Location for the output, together with protocol. This must be a writable
         directory.
+    consolidated : bool
+        Whether or not to add consolidated metadata key to references. (default True)
     storage_options: dict | None
         Passed to fsspec when for writing the parquet.
     row_group_size : int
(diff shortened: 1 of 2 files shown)


kerchunk/df.py
```diff
@@ -25,7 +28,182 @@ def _proc_raw(r):
...
+    compression : str
+        Compression information to pass to parquet engine, default is zstd.
+    engine : {'fastparquet', 'pyarrow'}
+        Library to use for writing parquet files.
+    **kwargs : dict
+        Additional keyword arguments passed to parquet engine of choice.
+    """
+    if not os.path.exists(url):
+        os.makedirs(url)
+    if "refs" in refs:
```
agoodm, line 122: Maybe there should be a better method to normalize refs here. The original `refs_to_dataframe` passes them into a separate `ReferenceFilesystem` to do this in one go, but I found that doing this is is really slow for large reference sets because (1) in the case a dict is passed, there is overhead from fsspec in tokenizing the input and (2) it needs to iterate through the whole reference set in case there are templates that need to be rendered. Wondering if we could/should add separate logic that skips doing this if there are no templates in the urls?

kerchunk/df.py
```diff
@@ -25,7 +28,182 @@ def _proc_raw(r):
...
+            )
+            if chunk_sizes.size == 0:
+                chunk_sizes = np.array([0])
+            nchunks = int(np.product(chunk_sizes))
+            extra_rows = row_group_size - nchunks % row_group_size
+            output_size = nchunks + extra_rows
+            # Initialize dataframe columns. Don't know why but int64 gives better
+            # compression ratio than int32
+            paths = np.full(output_size, np.nan, dtype="O")
+            offsets = np.zeros(output_size, dtype="int64")
```
agoodm, line 149: I noticed this in the comment but was wondering why I was getting better compression ratios setting this to `int64`, because it would definitely be ideal to use `int32` instead here.

kerchunk/df.py
```diff
@@ -25,7 +28,182 @@ def _proc_raw(r):
...
+                out_path,
+                engine=engine,
+                storage_options=storage_options,
+                compression=compression,
+                index=False,
+                **kwargs,
+            )
+
+
+def refs_to_dataframe_full(
```
agoodm, line 206: Wondering how the two functions should be named. I think it would be great to keep this version around for combining multiple reference sets since it should be quicker and simpler to use for that purpose.

agoodm (commented):
@martindurant When you are ready I have left a few additional comments on some specific sections that I would appreciate some additional feedback and clarity on.

kerchunk/df.py
martindurant: There could easily be chunks bigger than 4GB, let's leave it. For the compression ratio - I don't know, the algorithms are magic.

rsignell-usgs:
@martindurant where do we stand with this?

martindurant:
I have the fork fetched locally and I'm excited to play with it!

agoodm:
Just remembered something kind of important. Unlike the other approaches used for loading references, this one very strongly benefits from having a consolidated metadata key included. Would there be any objections to incorporating this?

```python
# inside refs_to_dataframe()
if consolidated: # optional arg (default True)
    refs = zarr.consolidate_metadata(refs)
```

martindurant:
> this one very strongly benefits from having a consolidated metadata

I suppose this is the real "zarr" way to do what I was trying to do with my metadata partition.

martindurant:
@agoodm , I made https://github.com/agoodm/kerchunk/pull/1 with suggestions instead of commenting on your code directly.

agoodm:
Thanks! I will have some time to take a closer look at this later tomorrow.

martindurant:
I updated the docstring formatting a little, and will merge when it goes green again. We can consider the preffs multi-component references as a followup, or when someone asks for the feature.

agoodm:
Thanks!

Actually I need to apologize for one small little niggle, I made that last commit by accident and meant to keep it in your branch, then merge all the changes to mine directly once we had finalized everything. I ended up finding one small little bug in the expected number of reference parquet files not being counted correctly and that commit was not included in this merge. I'll quickly make another PR with it so it can also be included.

PR #308, opened 2023-02-27
Fix bug in `combine.auto_dask`

### LINKED ISSUE #306: `combine.auto_dask` doesn't loop over all batches
I think there's an bug in the [batch loop in `combine.auto_dask`](https://github.com/fsspec/kerchunk/blob/e235ddd96cf2311f413612b20469af4aab29a75b/kerchunk/combine.py#L677)

Currently, it loops over 

```python
tasks_per_batch = len(tasks) // n_batches

for batch in range(tasks_per_batch + 1):
    in_tasks = tasks[batch * tasks_per_batch : (batch + 1) * tasks_per_batch]
    ...
```
This will skip batches/tasks when `n_batches` > `tasks_per_batch`. Should the loop not be something like:
```python
tasks_per_batch = -(-len(tasks) // n_batches)

for batch in range(n_batches):
    in_tasks = tasks[batch * tasks_per_batch : (batch + 1) * tasks_per_batch]
    ...
```

Happy to submit a PR if you agree.

dougiesquire:
Simple fix to a looping bug in `combine.auto_dask`. Includes modification of existing test that fails on pre-fix code.

Closes #306

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -672,9 +672,9 @@ def auto_dask(
 
     # make delayed calls
     tasks = [single_task(u) for u in urls]
-    tasks_per_batch = len(tasks) // n_batches
+    tasks_per_batch = -(-len(tasks) // n_batches)
     tasks2 = []
-    for batch in range(tasks_per_batch + 1):
+    for batch in range(n_batches):
         in_tasks = tasks[batch * tasks_per_batch : (batch + 1) * tasks_per_batch]
         u = urls[batch * tasks_per_batch : (batch + 1) * tasks_per_batch]
         if in_tasks:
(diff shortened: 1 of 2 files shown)


kerchunk/combine.py
```diff
@@ -672,9 +672,9 @@ class with ``translate()`` method
 
     # make delayed calls
     tasks = [single_task(u) for u in urls]
-    tasks_per_batch = len(tasks) // n_batches
+    tasks_per_batch = -(-len(tasks) // n_batches)
```
martindurant, line 675: why minus minus? This rounds up?

dougiesquire: Yup, I think for the way this loop is now written, we want to round up. E.g. for 3 tasks and 2 batches we'd want `tasks_per_batch` = 2

martindurant:
Failing test will be fixed by https://github.com/fsspec/filesystem_spec/pull/1197

PR #318, opened 2023-03-13
refs_to_dataframe() reference file count fix

agoodm:
@martindurant This adds a quick fix to `df.py` so that the total number of reference parquet files is now counted correctly per the changes introduced in #298. Without it the 2nd to last file may have an incomplete number of references included.


agoodm:
I think it should be good to go now, sorry about any confusion!

In case it is unclear what was fixed, you can see the change I made to `test_df.py` where one of the references was missing in the loaded dataframe such that the size was not consistent with the specified record_size.

agoodm:
Thanks!

PR #323, opened 2023-03-28
Support urls as references input to refs_to_dataframe

agoodm:
Tested with:

```python
url = 's3://esip/noaa/nwm/grid1km/LDAS_combined.json'
target_options = dict(anon=True, skip_instance_cache=True, 
                                     client_kwargs={'endpoint_url':'https://ncsa.osn.xsede.org'})
refs_to_dataframe(url, 'test_parq', target_protocol='s3', target_options=target_options)

### DIFF
--- a/kerchunk/df.py
+++ b/kerchunk/df.py
@@ -100,8 +100,10 @@ def _write_json(fname, json_obj, storage_options=None):
 
 
 def refs_to_dataframe(
-    refs,
+    fo,
     url,
+    target_protocol=None,
+    target_options=None,
     storage_options=None,
     record_size=10_000,
     categorical_threshold=10,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: kerchunk/df.py [hunk 1]
```diff
@@ -132,6 +139,18 @@ def refs_to_dataframe(
     **kwargs : dict
         Additional keyword arguments passed to fastparquet.
     """
+    # Taken from fsspec/implementations/reference.py
+    if isinstance(fo, str):
+        # JSON file
+        dic = dict(**(target_options or {}), protocol=target_protocol)
+        ref_fs, fo = fsspec.core.url_to_fs(fo, **dic)
+        with ref_fs.open(fo, "rb") as f:
```
martindurant, line 147: Unless we will need `ref_fs` again, this is a simpler version of the same
```suggestion
        with fsspec.open(fo, "rb", **dic) as f:
```

PR #340, opened 2023-07-11
Implement JSON embed for complex HDF5 arrays

### LINKED ISSUE #339: UserWarning / NotImplementedError
Attempting to index the following file (should be public)

```
s3://earthmover-sample-data/hdf5/NEONDSTowerTemperatureData.hdf5
```

with the following code

```python
from kerchunk.hdf import SingleHdf5ToZarr
import fsspec

s3_url = "s3://earthmover-sample-data/hdf5/NEONDSTowerTemperatureData.hdf5"

with fsspec.open(s3_url) as infile:
    h5chunks = SingleHdf5ToZarr(infile, s3_url, inline_threshold=0)
    out = h5chunks.translate()
```

produces a rather uninformative warning

```
/srv/conda/envs/notebook/lib/python3.10/site-packages/kerchunk/hdf.py:436: UserWarning: The following excepion was caught and quashed while traversing HDF5

Traceback (most recent call last):
  File "/srv/conda/envs/notebook/lib/python3.10/site-packages/kerchunk/hdf.py", line 363, in _translator
    raise NotImplementedError
NotImplementedError

  warnings.warn(msg)
```

Following the code

https://github.com/fsspec/kerchunk/blob/f272c66c9ea23dd8aaef885f9b6fb1c1312d3813/kerchunk/hdf.py#L362-L363

Brings me to https://github.com/zarr-developers/numcodecs/issues/333. The issue is closed, suggesting that the underlying problem is resolved. If so, why are we getting a `NotImplementedError`.

Can someone explain what is going on?

martindurant:
Fixes #339 

@rabernat , looks OK? This gives an Object array rather than recreating the complex dtype of the original, but that seems fine for JSON storage.


rabernat:
Thanks for working on this Martin.

I don't really understand what this code is doing. Does it put all of the data into the json file?

It seems fine to me for the example file, but I don't have a deep understanding of how this might go wrong.

martindurant:
> Does it put all of the data into the json file?

Yes, that what the "embed" option is for. It still uses numcodecs to format it, though (json within json), since zarr doesn't want to know about kerchunk, which is why it needed the upstream fix before I could implement it here.

I'll merge this for now, until someone comes along with an example that breaks it :)

Issue #342, opened 2023-08-03
translate error "ValueError: Shuffle buffer is not an integer multiple of elementsize"


kthyng:
Hi! I am going through the quick start to try out kerchunk (finally!). Unfortunately I hit what is to me a weird error and I am not sure what direction to go, being new to the library and the error. 

I'm following the steps from the Quick Start: https://fsspec.github.io/kerchunk/test_example.html#

I am using local netcdf files, not from S3. Perhaps this is the issue because all the examples I can find are for files on S3 instead of local files.

Here is my code, though it isn't reproducible since the files are local

```
import kerchunk.hdf
import fsspec
import pathlib
from kerchunk.combine import MultiZarrToZarr

base = pathlib.Path("/mnt/depot/data/packrat/prod/aoos/nwgoa/processed/")
urls = sorted(base.glob("1999/*1999-01*.nc"))

so = dict(
    # anon=True, 
    default_fill_cache=False, default_cache_type='first'
)
singles = []
for u in urls:
    with fsspec.open(u, **so) as inf:
        h5chunks = kerchunk.hdf.SingleHdf5ToZarr(inf, u, inline_threshold=100)
        singles.append(h5chunks.translate())

# Multi-file JSONs
mzz = MultiZarrToZarr(
    singles,
    # remote_protocol="s3",
    remote_options={'anon': True},
    concat_dims=["ocean_time"],
)

out = mzz.translate()
```

The error is  `ValueError: Shuffle buffer is not an integer multiple of elementsize` and here is the stack trace

<details>
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[30], line 9
      1 mzz = MultiZarrToZarr(
      2     singles,
      3     remote_protocol='file',
   (...)
      6     concat_dims=["ocean_time"],
      7 )
----> 9 out = mzz.translate()

File ~/miniconda3/envs/xp-inventory/lib/python3.11/site-packages/kerchunk/combine.py:496, in MultiZarrToZarr.translate(self, filename, storage_options)
    490 """Perform all stages and return the resultant references dict
    491 
    492 If filename and storage options are given, the output is written to this
    493 file using ujson and fsspec instead of being returned.
    494 """
    495 if 1 not in self.done:
--> 496     self.first_pass()
    497 if 2 not in self.done:
    498     self.store_coords()

[...]

PR #350, opened 2023-08-15
Basic support for HDF5 filters

florianziemen:
So far only blosc and zstd.
The mapping between hdf5 filters and numcodec filters is not trivial.


kerchunk/hdf.py
```diff
@@ -189,12 +189,40 @@ def _transfer_attrs(
...
+
+
+    decoders = { "32001" : _decode_blosc,
+                 "32015" : _decode_zstd,
+                }
+
+    def _decode_filters(self, h5obj: Union[h5py.Dataset, h5py.Group]):
+        if len(h5obj._filters.keys()) > 1:
+            raise RuntimeError(
+                f"{h5obj.name} uses multiple filters {list (h5obj._filters.keys())}. This is not supported by kerchunk."
```
martindurant, line 208: This came up in https://github.com/fsspec/kerchunk/issues/342 - actually zarr *does* support multiple filters, except only the final ne is called "compression"

martindurant:
What would you need to get unstuck here? We should be able to generate HDF files with various compressions even double-compression.

martindurant:
Merging this, thanks. Should support multiple filters, but this is not tested. It stuffs all filters into zarr's "filters" without calling any one of them the compression, which is fine.

PR #354, opened 2023-08-23
Ensure `inline_array` keeps attributes

jrbourbeau:
Currently `inline_array` won't keep attrs around for arrays that it inlines. This PR makes sure attrs are preserved. 

cc @martindurant

### DIFF
--- a/kerchunk/tests/test_utils.py
+++ b/kerchunk/tests/test_utils.py
@@ -70,12 +70,16 @@ def test_inline_array():
 """,
         "data/0": b"\x01\x00\x00\x00",
         "data/1": b"\x02\x00\x00\x00",
+        'data/.zattrs': '{"_ARRAY_DIMENSIONS":["Longitude","nv"],"comment":"longitude '
+                        'values at the west and east bounds of each '
+                        'pixel.","units":"degrees_east"}',
     }
     fs = fsspec.filesystem("reference", fo=refs)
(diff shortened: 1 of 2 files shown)


martindurant:
We are hitting a hard crash here (on shutdown of the pytest session), which I can only imagine is coming from HDF5 or maybe grib. This will take some debugging...

jrbourbeau:
Just checking, my guess is this failure is happening on `main` too. I don't see how that could be related to the changes here (though maybe I'm missing something)

martindurant:
> this failure is happening on main too

Not locally :|

jrbourbeau:
I just updated the test added here to use a more simple `attrs`. This seemed to help locally, let's see if it helps in CI. I don't know why the previous `attrs` were problematic, but the simplified `attrs` should still be sufficient

martindurant:
Since it's OK for py38 but crashed on py310, suggests a version problem. I note h5py 3.9.0 just came out.

martindurant:
Confirmed: pinning h5py makes tests pass https://github.com/fsspec/kerchunk/actions/runs/5978456412/job/16220596021?pr=349

martindurant:
^ cc @ajelenak

martindurant:
Pushing the temp fix here, and we can deal with whatever is the underlying problem separately.

jrbourbeau:
Thanks @martindurant

PR #359, opened 2023-09-20
Remove CI warnings

ianthomas23:
Remove warnings in CI:

- Replace deprecated `mamba-org/provision-with-micromamba` with `mamba-org/setup-micromamba`.
- Update versions of actions that are using deprecated `node12`.

### DIFF
--- a/.github/workflows/default.yml
+++ b/.github/workflows/default.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


ianthomas23:
Test failures look pretty fundamental. Maybe due to new version of NumPy (1.26.0)?

martindurant:
It just looks like a remote test file disappeared?

ianthomas23:
I switched to using a (randomly selected) different file on s3 from the same bucket and it looks OK now.

Issue #150, opened 2022-04-14
scan_grib is not outputting all the pieces dimensions correctly.


chiaral:
I am very new at this, so I might not comprehend the exact goal of `scan_grib` but here we go (this is a follow up from [discourse](https://discourse.pangeo.io/t/accessing-grib2-files-as-a-single-cloud-friendly-dataset-in-xarray-through-kerchunk/2401/12)):

**What's the issue**: `scan_grib` doesn't outputs all the "parts" that I expect to see in the json files, hence when i read the file through xarray I am missing parts of it. Something is wrong in the **writing** of the parsing of the file that comes out of `_split_file`. I explored a bit, but couldn't find what the issue is.

grib_file to download:
`!wget https://noaa-gefs-retrospective.s3.amazonaws.com/GEFSv12/reforecast/2000/2000010100/c00/Days:1-10/acpcp_sfc_2000010100_c00.grib2`

versions I am using:
<img width="219" alt="Screen Shot 2022-04-14 at 3 13 41 PM" src="https://user-images.githubusercontent.com/8453445/163460413-8434207e-928c-4125-915b-140fd73f239c.png">

```
import xarray as xr
import cfgrib

dsx = xr.open_dataset('acpcp_sfc_2000010100_c00.grib2', engine='cfgrib')
dscf = cfgrib.open_file('acpcp_sfc_2000010100_c00.grib2')
```
in both cases (which are essentially the same thing) the dimensions are recognized correctly, in particular:

`dsx`
<img width="743" alt="Screen Shot 2022-04-14 at 2 53 45 PM" src="https://user-images.githubusercontent.com/8453445/163457758-ea90a5fb-918f-4079-827f-ad3096546d47.png">

`dscf`
```
Dataset(
dimensions={'step': 80, 'latitude': 721, 'longitude': 1440}, 
variables={'number': Variable(dimensions=(), data=0), 
'time': Variable(dimensions=(), data=946684800), 
'step': Variable(dimensions=('step',), data=array([  3.,   6.,   9.,  12.,  15.,  18.,  21.,  24.,  27.,  30.,  33.,
                    ...condensed..for...clarity    
       201., 204., 207., 210., 213., 216., 219., 222., 225., 228., 231.,
       234., 237., 240.])), 
'surface': Variable(dimensions=(), data=0.0),
[...]

Issue #361, opened 2023-09-21
Represent (coordinate) variables "symbolically"


TomAugspurger:
I'm working with a GRIB2 file, and am interested in minimizing the size of the references file. Currently, the largest values in the references come from the base64-encoded coordinates that were inlined in the references:

```
  'latitude/0': 'base64:AAAAAACAVkBmZmZ...',
```

This specific variable (and longitude) can be represented "symbolically" (maybe not the right name), with something like a `range(90, -90.1, -0.4)`.

My questions:

1. Does something like this make sense to try?
2. Does this instead belong in Zarr instead? It seems more generally useful to compress the size of the data, beyond just what Kerchunk inlines (though I'd still want it in Kerchunk, so that inlined references can benefit from it).

Somewhat annoyingly, there are floating point inaccuracies between what I get from `np.arange` and what's coming out of cfgrib. But hopefully those can be solved.

martindurant:
This is certainly something that kerchunk could do, with effectively our own codec to expand whatever representation into an array at read time. That would be simple for linear coordinates, but GRIB allows for many complex coordinate definitions. I suppose it's possible to extract the parameters of whatever the coordinate system is, but we probably don't want to implement the coordinate generation algorithms, but  _call_ the appropriate functions in eccodes itself, if we can.

This all connects to the possibility of analytical coordinates in xarray. Perhaps we shouldn't be making arrays even at read time but making xarray indexes.

dcherian:
There's a [CF convention](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#appendix-coordinate-subsampling) for that!

We could totally interpret those as a "functional xarray index" too.

martindurant:
> There's a [CF convention](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#appendix-coordinate-subsampling) for that

(plus also the FITS WCS ways to define the same; you won't get these from geo-datasets, but I think they may be more general)

martindurant:
People on this thread might be interested in the intake-stac sprint https://github.com/intake/intake-stac/issues/159

TomAugspurger:
Thanks @dcherian. IIUC, the coordinate subsampling you linked to is essentially the same as `range(0, 10, 1)`? We just have two "tie points" (the first and last point) and then linearly interpolate between them?

[...]

PR #363, opened 2023-09-25
Improvements to scan_grib

dcherian:
xref #150 

- [x] needs tests

### DIFF
--- a/kerchunk/codecs.py
+++ b/kerchunk/codecs.py
@@ -106,7 +106,7 @@ class GRIBCodec(numcodecs.abc.Codec):
                 if out is not None:
                     return numcodecs.compat.ndarray_copy(data, out)
                 else:
-                    return data.astype(dt)
+                    return data.astype(dt, copy=False)
 
             finally:
                 eccodes.codes_release(mid)
--- a/kerchunk/codecs.py
+++ b/kerchunk/codecs.py
@@ -91,21 +91,23 @@ class GRIBCodec(numcodecs.abc.Codec):
 
         if self.var in ["latitude", "longitude"]:
             var = self.var + "s"
-            dt = self.dtype or "float64"
+            dtype = self.dtype or "float64"
         else:
             var = "values"
-            dt = self.dtype or "float32"
+            dtype = self.dtype or "float32"
+        if isinstance(dtype, str):
+            dtype = np.dtype(dtype).type
         with self.eclock:
             mid = eccodes.codes_new_from_message(bytes(buf))
             try:
-                data = eccodes.codes_get_array(mid, var)
+                data = eccodes.codes_get_array(mid, var, dtype)
                 missingValue = eccodes.codes_get_string(mid, "missingValue")
                 if var == "values" and missingValue:
                     data[data == float(missingValue)] = np.nan
                 if out is not None:
                     return numcodecs.compat.ndarray_copy(data, out)
                 else:
-                    return data.astype(dt, copy=False)
+                    return data
 
             finally:
                 eccodes.codes_release(mid)
--- a/kerchunk/codecs.py
+++ b/kerchunk/codecs.py
@@ -91,23 +91,21 @@ class GRIBCodec(numcodecs.abc.Codec):
 
         if self.var in ["latitude", "longitude"]:
             var = self.var + "s"
-            dtype = self.dtype or "float64"
+            dt = self.dtype or "float64"
         else:
             var = "values"
-            dtype = self.dtype or "float32"
-        if isinstance(dtype, str):
-            dtype = np.dtype(dtype).type
+            dt = self.dtype or "float32"
         with self.eclock:
             mid = eccodes.codes_new_from_message(bytes(buf))
             try:
-                data = eccodes.codes_get_array(mid, var, dtype)
+                data = eccodes.codes_get_array(mid, var)
                 missingValue = eccodes.codes_get_string(mid, "missingValue")
                 if var == "values" and missingValue:
                     data[data == float(missingValue)] = np.nan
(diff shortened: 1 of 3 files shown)


martindurant:
Please let me know when you are ready. I must admit to being largely in the dark on this one.

dcherian:
> being largely in the dark on this one

me too! I'm trying to minimize differences with cfgrib which meant
- keeping coordinates as scalars
- lots of minor changes to attributes (we could just avoid this too...)

I see that kerhcunk always tries to inline the dimension coordinate variables. Is there a way to avoid that and use the byte range instead?

martindurant:
>  Is there a way to avoid that and use the byte range instead?

Coordinates are included in messages; so the only way to get the coordinates is to read the whole message once each for each coordinate (this is how zarr calls). Whereas the payload of a message is presumably the largest component, the coordinates are represented analytically internally - so we have no buffer to read, but we could in theory generate our own from the parameters alone rather than inlining ( #361 ).

kerchunk/grib2.py
```diff
@@ -190,12 +246,12 @@ def scan_grib(
                 coord2 = {"latitude": "latitudes", "longitude": "longitudes"}.get(
                     coord, coord
                 )
-                if coord2 in m:
-                    x = m[coord2]
-                else:
+                x = m.get(coord2)
+                if x is None:
                     continue
```
dcherian, line 251: This is a large improvement because even `__contains__` reads out "coord2".

Unfortunately, there doesn't seem to be any way to get a list of valid keys for a message (see comments above).

Even cfgrib just tries indexing and catches the error:
https://github.com/ecmwf/cfgrib/blob/f9106c8bc6b18238aa1e7d54100e6e6364d62ffe/cfgrib/dataset.py#L470-L479

kerchunk/grib2.py
```diff
@@ -149,33 +185,53 @@ def scan_grib(
...
             z.attrs.update(global_attrs)
 
-            vals = m["values"].reshape((m["Ny"], m["Nx"]))
+            if data_size < inline_threshold:
+                # read the data
+                vals = m["values"].reshape(shape)
+            else:
+                # dummy array to match the required interface
+                vals = np.empty(shape, dtype=native_type)
+                assert vals.size == data_size
```
dcherian, line 203: major improvement here.

dcherian:
I'll add some tests but I could use a surface review here. I've added a number of extra comments, that I'm happy to delete if you would like.

kerchunk/grib2.py
```diff
@@ -149,33 +185,53 @@ def scan_grib(
...
+                    attrs[k] = attrs[v]
+
             # try to use cfVarName if available,
             # otherwise use the grib shortName
             varName = m["cfVarName"]
             if varName in ("undef", "unknown"):
                 varName = m["shortName"]
             _store_array(store, z, vals, varName, inline_threshold, offset, size, attrs)
-            if "typeOfLevel" in m and "level" in m:
+            if "typeOfLevel" in message_keys and "level" in message_keys:
```
dcherian, line 223: small optimization, could be removed.

martindurant:
As far as I can tell, your explanations and comments seem to make sense. I wonder if we can approach some GRIB sage as a group and try to get involvement - perhaps ping devs of eccodes and cfgrib directly? Aside from the optimisations here, there are a whole long list of requests we would like to make, first to mind:
- extracting the coordinate parameters, and finding/calling the algos directly
- getting the payload buffer byte range and encoding for a given message

@dcherian , have you glanced kerchunk.codecs.GRIBCodec at all? In the "scan once, read many times" model, any improvement there would have a much bigger impact.

dcherian:
> a whole long list of requests we would like to make

How about you make an issue and I'll ping the usernames I know. (unfortunately they tend to respond slowly)

> have you glanced kerchunk.codecs.GRIBCodec at all? In the "scan once, read many times" model, any improvement there would have a much bigger impact.

Not yet. my task at hand is to make the references. I'll take a look and see.

dcherian:
Can we review+merge+release here please?

kerchunk/codecs.py
```diff
@@ -106,7 +106,7 @@ def decode(self, buf, out=None):
                 if out is not None:
                     return numcodecs.compat.ndarray_copy(data, out)
                 else:
-                    return data.astype(dt)
+                    return data.astype(dt, copy=False)
```
dcherian, line 109: small 5% improvement. 81% of the time is now in `eccodes.codes_get_array`, I'm not sure theres any improvement to be had there.

martindurant: > 81% of the time is now in eccodes.codes_get_array, I'm not sure theres any improvement to be had there.

👍 

I don't think there's anything we can do without delving deep into the low-level codebase.

dcherian: `codes_get_array` does take a `ktype` so technically we could push the casting there. I'm not sure if it supports all the dtypes though.

[API doc](https://sites.ecmwf.int/docs/eccodes/namespaceec_codes.html#a648bae4beadb6a82698da40e08323be9)
```
ktype | the type we want the output in (can be int, float or string), native type if not specified
```

martindurant: Worth a try?

I have had a delve of the grib .cc code hoping to isolate the buffer find and decode and most admit to being lost in the many layers of obfuscation.

dcherian: Pushed a commit, there's no real difference on my test case because the dtypes are the same.

martindurant:
The latest change seems to have broken tests

martindurant:
I am going to merge this now, and the we shall see if anything breaks for others :)

PR #366, opened 2023-09-29
Remove `breakpoint()` from hdf

norlandrhagen:
Seems like this is a leftover from some pdb debugging?


martindurant:
I suppose so - although we retain the possibility of using the debugger on demand for failing hdf chunks.

Issue #365, opened 2023-09-29
NetCDF3ToZarr: `KeyError: '__padding_16' when translating


mpiannucci:
To Reproduce: 

Download https://noaa-ofs-pds.s3.amazonaws.com/ngofs2.20230929/nos.ngofs2.2ds.f001.20230929.t15z.nc , then: 

```python
from kerchunk.netCDF3 import NetCDF3ToZarr

chunks = NetCDF3ToZarr('./nos.ngofs2.2ds.f001.20230929.t15z.nc')
translated = chunks.translate()
```

This yields: 

```
File [~/Developer/ocean-notebooks/env/lib/python3.10/site-packages/kerchunk/netCDF3.py:214](https://file+.vscode-resource.vscode-cdn.net/Users/matthewiannucci/Developer/ocean-notebooks/~/Developer/ocean-notebooks/env/lib/python3.10/site-packages/kerchunk/netCDF3.py:214), in NetCDF3ToZarr.translate(self)
    211 offset = start
    212 for name in dt.names:
    213     # the order of the names if fixed and important!
--> 214     var = self.variables[name]
    215     dtype = dt[name]
    216     base = dtype.base  # actual dtype

KeyError: '_padding_16'
```

in `netcdf3.py` changing this on line 210 seems make it work fine but i am not sure if this breaks anything. Im not sure the impact the padding has. 

```python
        if "record_array" in self.chunks:
            # native chunks version (no codec, no options)
            start, size, dt = self.chunks["record_array"][0]
            dt = np.dtype(dt)
            outer_shape = size // dt.itemsize
            offset = start
            for name in dt.names:
 +               if name not in variables: 
 +                  continue
```

Any hints would help, happy to make PR if necessary

mpiannucci:
Fixed with #367

PR #367, opened 2023-10-02
Handle padding in netcdf3 translation

mpiannucci:
#365 

The padding was not being handled, so the translation crashed because it was not an actual variable. This PR skips the padding but makes sure to increment the offset for each padding chunk encountered.

### DIFF
--- a/kerchunk/netCDF3.py
+++ b/kerchunk/netCDF3.py
@@ -207,9 +207,15 @@ class NetCDF3ToZarr(netcdf_file):
             outer_shape = size // dt.itemsize
             offset = start
             for name in dt.names:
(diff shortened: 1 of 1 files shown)


martindurant:
Can we make a file with such padding for the sake of tests?

mpiannucci:
I can try and cut the one i have down maybe, this is the first time i have ever encountered it and the example is pretty large

mpiannucci:
Added a simple test case to the existing unlimited test. Without the padding being handled as it is in this PR, the test will fail because the temperature offset will be incorrect.

martindurant:
This is perfect

Issue #368, opened 2023-10-04
Coordinate Variables have incorrect shapes and chunks if name matches a dimension


mpiannucci:
FVCOM model data has a quirk where the `siglay` and `siglev` coordinate names match the `siglay` and `siglev` dimensions (See https://github.com/pydata/xarray/issues/2233). XArray can now handle this but kerchunk wrongly assumes that if a variable matches the name of a dimension then it must have a single dimension matching its name. 

Here is an example FVCOM dataset: https://noaa-ofs-pds.s3.amazonaws.com/ngofs2.20231003/nos.ngofs2.fields.f042.20231003.t09z.nc

Loading this in gives us the following for `siglay` and `siglev`:

![Screenshot 2023-10-04 at 10 00 55 AM](https://github.com/fsspec/kerchunk/assets/1832671/bde4c662-94e3-482c-8c7a-27eb311a50e6)

Note that the `siglay` dimension is `40`, `siglev` is `41`, but the `siglay` variable has a shape of `(40, 303714)` and `siglev` has a shape of `(41, 303714)`. 

When we process this dataset with kerchunk the result is that the dimensions match but the shapes have become `(40)` and `(41)` respectively: 

```json
        "siglay\/.zarray": "{\"chunks\":[40],\"compressor\":null,\"dtype\":\">f4\",\"fill_value\":null,\"filters\":null,\"order\":\"C\",\"shape\":[40],\"zarr_format\":2}",
        "siglay\/0": [
            ".\/nos.ngofs2.fields.f042.20231003.t09z.nc",
            16258148,
            48594240
        ],
        "siglay\/.zattrs": "{\"_ARRAY_DIMENSIONS\":[\"siglay\",\"node\"],\"formula_terms\":\"sigma: siglay eta: zeta depth: h\",\"long_name\":\"Sigma Layers\",\"positive\":\"up\",\"standard_name\":\"ocean_sigma\\\/general_coordinate\",\"valid_max\":\"0.0\",\"valid_min\":\"-1.0\"}",
        "siglev\/.zarray": "{\"chunks\":[41],\"compressor\":null,\"dtype\":\">f4\",\"fill_value\":null,\"filters\":null,\"order\":\"C\",\"shape\":[41],\"zarr_format\":2}",
        "siglev\/0": [
            ".\/nos.ngofs2.fields.f042.20231003.t09z.nc",
            64852388,
            49809096
        ],
[...]

PR #369, opened 2023-10-04
Handle matching coordinates and dimensions with differing shapes

mpiannucci:
#368 

This seems to be specific to 64bit netcdf, so I added a new test case for that specific case.


martindurant:
Perhaps the test requires a certain version of scipy/netcdf/xarray? I know not all packages are still releasing for py38.

mpiannucci:
> Perhaps the test requires a certain version of scipy/netcdf/xarray? I know not all packages are still releasing for py38.

Yeah it requires XArray 2023.8.0. Do you know the best way to handle that?

martindurant:
Skip the test on py38, or test the xarray version

Issue #360, opened 2023-09-21
Add `xarray.open_dataset` backend


TomAugspurger:
https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html describes how to add a new backend to xarray's open_dataset. Adding a backend will make it easier for users to consume kerchunk references.

Here's a prototype:

```python
from xarray.backends import BackendEntrypoint


class KerchunkBackendEntrypoint(BackendEntrypoint):
    def open_dataset(
        self,
        filename_or_obj,
        *,
        drop_variables=None,
        **kwargs,
    ):
        return my_open_dataset(filename_or_obj, drop_variables=drop_variables)

    open_dataset_parameters = ["filename_or_obj", "drop_variables"]

    def guess_can_open(self, filename_or_obj):
        try:
            _, ext = os.path.splitext(filename_or_obj)
        except TypeError:
            return False
        return ext in {".json", ".parquet"}

    description = "Use Kerchunk files in Xarray"
    url = "https://fsspec.github.io/kerchunk/"

def my_open_dataset(filename_or_obj, drop_variables, **kwargs):
    m = fsspec.filesystem("reference", fo=filename_or_obj, **kwargs).get_mapper()
    return xr.open_dataset(m, engine="zarr", consolidated=False)
```

This would be usable like

```python
ds = xr.open_dataset(refs, engine="kerchunk")
ds
```

martindurant:
Is it possible to implement this as an entrypoints plugin, or does it need to be a PR to xarray?

The only trouble I see, is that the definition of an open process for kerchunk can be quite complex, so we might have a lot of arguments now inside of backend_kwargs. Whilst engine="kerchunk" might be convenient in some situations, In the end these definitions should probably live in catalogs.


```
     return ext in {".json", ".parquet"}
```
The parquet storage is a directory, which I suppose should have "parquet" in the name. We should probably check for the .zmetadata file inside the root.

TomAugspurger:
> Is it possible to implement this as an entrypoints plugin, or does it need to be a PR to xarray?

As a plugin, I believe.

> In the end these definitions should probably live in catalogs.

Probably, though you need to specify them either way.

The main benefit this brings is the user not needing to deal with the reference file system stuff like .

```
ds = xr.open_dataset("reference://", engine="zarr", ...)
```

or

```
fsspec.filesystem("reference", fo=filename_or_obj, **kwargs).get_mapper()
```

[...]

PR #372, opened 2023-10-06
Xarray <-> Kerchunk backend

norlandrhagen:
Based off this [issue](https://github.com/fsspec/kerchunk/issues/360#issuecomment-1751389780) opened by @TomAugspurger and the [Pangeo Discourse post](https://discourse.pangeo.io/t/making-kerchunk-as-simple-as-a-toggle/3690) opened by @ahuang11.

```
import xarray as xr
import kerchunk

ds = xr.open_dataset(<path_to_ref_file>, engine="kerchunk")


```


kerchunk/xarray_backend.py
```diff
@@ -0,0 +1,31 @@
...
+            _, ext = os.path.splitext(filename_or_obj)
+        except TypeError:
+            return False
+        return ext in {".json", ".parquet"}
+
+    description = "Open Kerchunk References with Xarray"
+    url = "https://fsspec.github.io/kerchunk/"
+
+
+def my_open_dataset(filename_or_obj, **kwargs):
```
TomAugspurger, line 28: Probably want this to have a meaningful name.

And I suspect we'll want to forward keyword arguments to two places, so we'll need `reference_fs_options` (or `storage_options`?) and `open_dataset_options` instead of `**kwargs`.

TomAugspurger (commented):
Thanks. It'd be good to add a basic test too.

kerchunk/xarray_backend.py
```diff
@@ -0,0 +1,31 @@
...
+import os
+from fsspec.implementations.reference import ReferenceFileSystem
+
+
+class KerchunkBackend(BackendEntrypoint):
+    def open_dataset(
+        self,
+        filename_or_obj,
+        **kwargs,
+    ):
```
ahuang11, line 12: Should there be a docstring?

kerchunk/xarray_backend.py
```diff
@@ -0,0 +1,31 @@
...
+        except TypeError:
+            return False
+        return ext in {".json", ".parquet"}
+
+    description = "Open Kerchunk References with Xarray"
+    url = "https://fsspec.github.io/kerchunk/"
+
+
+def my_open_dataset(filename_or_obj, **kwargs):
+    fs = ReferenceFileSystem(fo=filename_or_obj, **kwargs)
```
ahuang11, line 29: Should kwargs go here or `open_dataset`?

ahuang11 (commented):
Awesome!! Thanks for creating this; I think it'll be immensely helpful

norlandrhagen:
Np! Feel free to add, suggest or give it a try. Credit to @TomAugspurger for outlining it all.

kerchunk/xarray_backend.py
keewis: I'd probably have a signature of
```python
def open_reference_dataset(filename_or_obj, storage_options, open_dataset_kwargs):
    pass
```
(note, no `**kwargs`, since those will be caught by the backend entrypoint's `open_dataset`)

If you go with `storage_options`, you might have to disallow passing `fo` in there. If not, you could make `target_protocol`, `target_options`, `remote_protocol` and `remote_options` parameters (or combine them somehow into `target_options` and `remote_options`, not sure). Whatever you decide, `open_dataset_parameters` should list all the options for the backend.

kerchunk/xarray_backend.py
keewis: We currently don't have guidelines on how to best document entrypoints. It would be good to figure those out (and maybe standardize in a sphinx extension), but for now you will most likely only see the docstring when reading the code or when inspecting the entrypoint before directly passing it to `xr.open_dataset`

setup.py
```diff
@@ -24,6 +24,7 @@
         "netcdf3": ["scipy"],
     },
     entry_points={
+        "xarray.backends": ["kerchunk=kerchunk.xarray_backend:KerchunkBackend"],
```
keewis, line 27: no need to add the entrypoint in both `setup.cfg` and `setup.py` (or rather, it's probably better not to do that).

As a side note, I'd also recommend transitioning to `pyproject.toml` since that's the most stable way going forward (but that would definitely have to be a separate PR).

norlandrhagen:
Thanks for all the feedback @martindurant, @TomAugspurger , @ahuang11 and @keewis!

setup.py
martindurant: Would be happy to see a PR.

TomAugspurger (approved):

martindurant:
Let's give it a spin!

PR #379, opened 2023-10-19
Use pyproject toml and importlib.metadata

maxrjones:
Follows up on the recommendation in https://github.com/fsspec/kerchunk/pull/372#discussion_r1350070990 to use pyproject.toml rather than setup.py + setup.cfg. 

Still need to check whether there's any impacts from the deprecation of `cmdclass` and switch from versioneer to importlib.metadata.


martindurant:
Do you intend to do more here before taking off "WIP"?

maxrjones:
> Do you intend to do more here before taking off "WIP"?

Yes, want to double check on this:

> check whether there's any impacts from the deprecation of cmdclass and switch from versioneer to importlib.metadata.

I'll ping for a review after getting to that (should be tomorrow morning)

martindurant:
I would like to release today or this weekend, so it would be nice to make use of this work

maxrjones:
> I would like to release today or this weekend, so it would be nice to make use of this work

Looking now

maxrjones:
@martindurant the version format from setuptools_scm (e.g., `0.2.0.post62`) differs from the prior versioneer setup (e.g., `0.2.0+59.gff2d639`). The options from setuptools_scm include guessing the next minor version and appending the commit hash. Please let me know if you'd rather a different format. Otherwise, the passing tests and a quick check on the version suggests this is good to go.

martindurant:
I think that's fine.

PR #381, opened 2023-10-19
Adds check to ensure `consolidated=True` is not passed to open_dataset

norlandrhagen:
This pr is an tiny addition to https://github.com/fsspec/kerchunk/pull/372#event-10713351822.

### DIFF
--- a/kerchunk/xarray_backend.py
+++ b/kerchunk/xarray_backend.py
@@ -56,4 +56,4 @@ def open_reference_dataset(
 
     m = fsspec.get_mapper("reference://", fo=filename_or_obj, **storage_options)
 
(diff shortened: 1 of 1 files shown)


kerchunk/xarray_backend.py
```diff
@@ -48,6 +48,12 @@ def open_reference_dataset(
     if open_dataset_options is None:
         open_dataset_options = {}
 
+    if open_dataset_options.get("consolidated"):
+        raise ValueError(
```
martindurant, line 52: A warning would do? Without making this check, it only results in a warning from the zarr engine anyway I think. Note that parquet references do work as consolidated or not, it doesn't matter.

norlandrhagen: Ah good to know! Happy to close the PR.

PR #391, opened 2023-11-07
Direct Parquet in netCDF3 and HDF

martindurant:


kerchunk/combine.py
```diff
@@ -107,6 +107,7 @@ def __init__(
         preprocess=None,
         postprocess=None,
         out=None,
+        append=False,
```
NikosAlexandris, line 110: @martindurant Missing a :param append: entry in the docstring ?

martindurant: Yep, I had intended this to be just the start of a bigger PR, but since it's useful already, I'll tidy it up and remove this for a separate PR

Issue #313, opened 2023-03-06
Issue using `tiff_to_zarr`


norlandrhagen:
Hi there @martindurant,

I've started playing around with Kerchunk's `tiff_to_zarr` functionality and ran into an issue when trying to open up the reference file. In short, the `tiff_to_zarr` works successfully, but I get the error: `ContainsArrayError: path '' contains an array`. Hopefully this is a simple user error! 
I've successfully used `tiff_to_zarr` on the `kerchunk/tests/lcmap_tiny_cog_2019.tif` and this [Arctic DEM ](https://registry.opendata.aws/pgc-arcticdem/) without incident.

Thanks in advance!

<p>
</p>

I've included a code snippet below and a example of the generated reference file.

**Generation script:**

```python


import fsspec
import xarray as xr
import ujson
import imagecodecs.numcodecs
imagecodecs.numcodecs.register_codecs()

from kerchunk.tiff import tiff_to_zarr

hansen_url = 'https://storage.googleapis.com/earthenginepartners-hansen/GFC-2021-v1.9/Hansen_GFC-2021-v1.9_gain_50N_130W.tif'
ex_path = 'https://github.com/fsspec/kerchunk/blob/main/kerchunk/tests/lcmap_tiny_cog_2019.tif?raw=true'

hansen_ds = xr.open_dataset(hansen_url,engine='rasterio')


ex_ds = xr.open_dataset(ex_path,engine='rasterio')



zt = tiff_to_zarr(hansen_url)
with open("hansen.json", "wb") as f:
    f.write(ujson.dumps(zt).encode())

m = fsspec.get_mapper("reference://", fo='hansen.json')
hansen_kerchunk_ds = xr.open_dataset(
    m, engine="zarr", backend_kwargs={"consolidated": False}
)

```

examples of refs for both datasets:

**Hansen:**
```json
{
[...]

Issue #114, opened 2021-12-30
Update the required dependencies


TomAugspurger:
What are the required dependencies of kerchunk? Right now https://github.com/fsspec/kerchunk/blob/6b57f0620c9e1e55206523ac54dadf60830b0935/requirements.txt has ujson on fsspec.

Working with HDF5 files (https://github.com/fsspec/kerchunk/blob/main/kerchunk/hdf.py) seems to require h5py, zarr (and numcodecs), and numpy.

combine requires those plus xarray  and dask (since it uses `chunks={}`).

So I'd recommend the additional requirements

- h5py
- zarr
- numcodecs
- numpy
- xarray
- dask[array]

martindurant:
It probably makes sense to have the official deps be more comprehensive indeed. The original idea was, that you should be able to get the offsets for some grib or CSV without needing the rest of the things - but it's easy enough to install everything, given that makers of references will be far fewer than consumers. For now, at least, combined HDF5s will continue to be the most used path.

btw: I don't suppose you need numcodecs and numpy in the list, they are already included in the others.

PR #398, opened 2023-11-17
add deps for tests

emfdavid:
Add all dependencies for testing to the dev group.
Is there a better way to install all?

This should allow an update to the docs for [running the test suite ](https://fsspec.github.io/kerchunk/contributing.html#running-the-test-suite)
```console
pip install .[dev]
pytest kerchunk
```

Should `pip install -e .` install all these dependencies? Is this because I am using venv (not anaconda :cry:)?
Without explicitly installing the dev dependencies I don't have pytest and without adding the other dependencies (there are currently optional dependency groups for all but netcdf4), I can run the test suite.

Examples
```
_______________________________________________________________________ ERROR collecting external/kerchunk/kerchunk/tests/test_xarray_backend.py ________________________________________________________________________
ImportError while importing test module '/home/builder/bando/external/kerchunk/kerchunk/tests/test_xarray_backend.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
venv/lib/python3.10/site-packages/kerchunk/netCDF3.py:12: in <module>
    from scipy.io._netcdf import ZERO, NC_VARIABLE, netcdf_file, netcdf_variable
E   ModuleNotFoundError: No module named 'scipy'

During handling of the above exception, another exception occurred:
/usr/local/lib/python3.10/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
kerchunk/tests/test_xarray_backend.py:3: in <module>
    from kerchunk import netCDF3
venv/lib/python3.10/site-packages/kerchunk/netCDF3.py:14: in <module>
    raise ImportError(
E   ImportError: Scipy is required for kerchunking NetCDF3 files. Please install with `pip/conda install scipy`. See https://scipy.org/install/ for more details.
```

```
    @pytest.fixture()
    def matching_coordinate_dimension_dataset(tmpdir):
        """Create a dataset with a coordinate dimension that matches the name of a
        variable dimension."""
        # https://unidata.github.io/netcdf4-python/#creatingopeningclosing-a-netcdf-file
>       from netCDF4 import Dataset
E       ModuleNotFoundError: No module named 'netCDF4'

kerchunk/tests/test_netcdf.py:104: ModuleNotFoundError
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,9 @@ dev = [
     "s3fs",
     "types-ujson",
     "xarray",
+    "cfgrib",
+    "scipy",
+    "netcdf4"
 ]
 
 [project.urls]


maxrjones:
I actually ran into the same issue today when working on https://github.com/fsspec/kerchunk/issues/313. I got all of the tests to run using only pip with the following in `pyproject.toml` (which also moves dask based on https://github.com/fsspec/kerchunk/issues/114):

```
dependencies = [
  "fsspec",
  "dask[array]",
  "numcodecs<=0.11",
  "numpy",
  "ujson",
  "zarr",
]

[project.optional-dependencies]
cftime = ["cftime"]
fits = ["xarray"]
hdf = ["h5py", "xarray"]
grib2 = ["ecmwflibs", "eccodes", "cfgrib"]
netcdf3 = ["scipy"]
tiff = ["tifffile"]
all = ["kerchunk[cftime,fits,hdf,grib2,netcdf3,tiff]"]
dev = [
    "kerchunk[cftime,fits,hdf,grib2,netcdf3,tiff]",
    "astropy",
    "fastparquet",
    "h5netcdf",
    "ipfsspec",
    "jinja2",
    "mypy",
    "netCDF4",
    "pytest",
    "rioxarray",
    "s3fs",
    "types-ujson",
]
```

`kerchunk/tests/test_grib.py::test_one` failed for me though. Do you have `eccodes` installed separately? 

> Should pip install -e . install all these dependencies? 

Some of these only seem to be used in the tests (e.g., netCDF4)

martindurant:
> Should pip install -e . install all these dependencies? 

This should install the same packages as without `-e`. 

There was an issue arguing that we should include all the optional packages in a normal install, but kerchunk (the package) might be used not just for scanning files, but for the codecs needed during read, so I would rather not.

martindurant:
> Do you have eccodes installed separately?

cfgrib should bring that in I think

emfdavid:
Sorry for being thick - I don't work in anything like a normal python dependency / virtual env (Camus uses Bazel)
When I run `pip install -e .` I only get the following deps: `Successfully installed asciitree-0.3.3 entrypoints-0.4 fasteners-0.19 fsspec-2023.10.0 kerchunk-0.2.2.post12 numcodecs-0.11.0 numpy-1.26.2 ujson-5.8.0 zarr-2.16.1` because I am using python vritualenv, not anaconda.

The deps in the conda development instructions come from conda via the `ci/environment-py3<*>.yml` file. Does this have a corollary in the pip world?

I think just adding the same requirements from the ci/environment-py3 file to the [dev] group would do the trick. I agree we should keep the base requirements as lean as possible.

martindurant:
> When I run pip install -e . I only get the following deps

Yes, it's expected that the "editable" install (no longer called dev) should only get you the minimum. You would need to do
```
pip install -e .[dev]
```
to get the extras.

> Does this have a corollary in the pip world?

Some projects have a requirements-dev.txt, but requirements.txt is generally going away with the advent of pyproject.toml... The environments in ci/ are what we actually use for automated testing, so will give the closest experience, and it would be annoying to have to keep another package listing in sync with it. I am open to ideas!

martindurant:
In any case, this PR is useful, so merging it now.

PR #399, opened 2023-11-24
Add grib_tree method

emfdavid:
# Add an method to map grib2 data model to zarr hierarchy & xarray datatree

The method is opinionated about how to do this - there are many possible ways. I hope the is generally useful enough that it is worth adding to the kerchunk library.

See examples in [Colab](https://colab.research.google.com/drive/1fhNZYF5vhMBD7vuK9Kgmt50IUr_77kZE?usp=sharing)

Here is an example of the hierarchical structure for the eastward wind velocity variable which has been aggregated by step, time and level (isobaricInhPa).

Screenshot 2023-11-29 at 11 19 14 PM

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -311,8 +311,13 @@ class MultiZarrToZarr:
                         for _ in v
                     ]
                 ).ravel()
-            if "fill_value" not in kw and data.dtype.kind == "i":
-                kw["fill_value"] = None
+            if "fill_value" not in kw:
+                if data.dtype.kind == "i":
+                    kw["fill_value"] = None
+                elif k in z:
+                    # Fall back to existing fill value
+                    kw["fill_value"] = z[k].fill_value
+
             arr = group.create_dataset(
                 name=k,
                 data=data,
@@ -323,6 +328,9 @@ class MultiZarrToZarr:
             )
             if k in z:
                 # copy attributes if values came from an original variable
+                logger.warning(
+                    "Zarr attrs: %s", {ii: vv for ii, vv in z[k].attrs.items()}
+                )
                 arr.attrs.update(z[k].attrs)
             arr.attrs["_ARRAY_DIMENSIONS"] = [k]
             if self.cf_units and k in self.cf_units:
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -328,9 +328,6 @@ class MultiZarrToZarr:
             )
             if k in z:
                 # copy attributes if values came from an original variable
-                logger.warning(
-                    "Zarr attrs: %s", {ii: vv for ii, vv in z[k].attrs.items()}
-                )
                 arr.attrs.update(z[k].attrs)
             arr.attrs["_ARRAY_DIMENSIONS"] = [k]
             if self.cf_units and k in self.cf_units:
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -396,15 +396,13 @@ def grib_tree(
     # TODO allow passing a LazyReferenceMapper as output?
     zarr_store = {}
     zroot = zarr.open_group(store=zarr_store)
-    result = dict(refs=zarr_store)
 
     aggregations: Dict[str, List] = defaultdict(list)
     aggregation_dims: Dict[str, Set] = defaultdict(set)
 
     unknown_counter = 0
     for msg_ind, group in enumerate(message_groups):
-        if "version" not in result:
-            result["version"] = group["version"]
+        assert group["version"] == 1
 
         gattrs = ujson.loads(group["refs"][".zattrs"])
         coordinates = gattrs["coordinates"].split(" ")
@@ -517,6 +515,16 @@ def grib_tree(
         for key, value in group["refs"].items():
             if key not in [".zattrs", ".zgroup"]:
                 zarr_store[f"{path}/{key}"] = value
+
+    # Force all stored values to decode as string, not bytes. String should be correct.
+    # ujson will reject bytes values by default.
+    # Using 'reject_bytes=False' one write would fail an equality check on read.
+    zarr_store = {
+        key: (val.decode() if isinstance(val, bytes) else val)
+        for key, val in zarr_store.items()
+    }
+    # TODO handle other kerchunk reference spec versions?
+    result = dict(refs=zarr_store, version=1)
 
     return result
 
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -366,7 +366,9 @@ def grib_tree(
     remote_options=None,
 ) -> Dict:
     """
-    Build a hierarchical data model from a set of scanned grib messages. The iterable input groups should
+    Build a hierarchical data model from a set of scanned grib messages. 
+    
+    The iterable input groups should
     be a collection of results from scan_grib. Multiple grib files can be processed together to produce an
     FMRC like collection.
     The time (reference_time) and step coordinates will be used as concat_dims in the MultiZarrToZarr
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -396,6 +396,7 @@ def grib_tree(
     # Hard code the filters in the correct order for the group hierarchy
     filters = ["stepType", "typeOfLevel"]
 
+    # TODO allow passing a LazyReferenceMapper as output?
     zarr_store = {}
     zroot = zarr.open_group(store=zarr_store)
     result = dict(refs=zarr_store)
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -526,7 +526,9 @@ def grib_tree(
 
 def correct_hrrr_subhf_step(group: Dict) -> Dict:
     """
-    Overrides the definition of the step variable. Sets the value equal to the `valid_time - time`
+    Overrides the definition of the "step" variable. 
+    
+    Sets the value equal to the `valid_time - time`
     in hours as a floating point value. This fixes issues with the HRRR SubHF grib2 step as read by
     cfgrib via scan_grib.
     The result is a deep copy, the original data is unmodified.
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -379,9 +379,16 @@ def grib_tree(
     Grib steps that are missing due to WrongStepUnitError are patched with NaT
     The input message_groups should not be modified by this method
 
-    :param message_groups: a collection of zarr store like dictionaries as produced by scan_grib
-    :param remote_options: remote options to pass to ZarrToMultiZarr
-    :return: A new zarr store like dictionary for use as a reference filesystem mapper with zarr
+    Parameters
+    ----------
+    message_groups: iterable[dict]
+        a collection of zarr store like dictionaries as produced by scan_grib
+    remote_options: dict
+        remote options to pass to ZarrToMultiZarr
+
+    Returns
+    -------
+    list(dict): A new zarr store like dictionary for use as a reference filesystem mapper with zarr
     or xarray datatree
     """
     from kerchunk.combine import MultiZarrToZarr
@@ -516,14 +523,22 @@ def grib_tree(
     return result
 
 
-def correct_hrrr_subhf_step(group: dict) -> dict:
+def correct_hrrr_subhf_step(group: Dict) -> Dict:
     """
     Overrides the definition of the step variable. Sets the value equal to the `valid_time - time`
     in hours as a floating point value. This fixes issues with the HRRR SubHF grib2 step as read by
     cfgrib via scan_grib.
     The result is a deep copy, the original data is unmodified.
-    :param group: the zarr group store for a single grib message
-    :return: a new deep copy of the corrected zarr group store
+
+    Parameters
+    ----------
+    group: dict
+        the zarr group store for a single grib message
+
+    Returns
+    -------
+    dict: A new zarr store like dictionary for use as a reference filesystem mapper with zarr
+    or xarray datatree
     """
     group = copy.deepcopy(group)
     group["refs"]["step/.zarray"] = (
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -18,6 +18,7 @@ except ModuleNotFoundError as err:  # pragma: no cover
 
 import fsspec
 import zarr
+import xarray
 import numpy as np
 
 from kerchunk.utils import class_factory, _encode_for_JSON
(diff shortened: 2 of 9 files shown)


kerchunk/grib2.py
```diff
@@ -354,3 +359,197 @@ def example_combine(
...
     )
     return mzz.translate()
+
+
+def grib_tree(
+    message_groups: Iterable[Dict],
+    remote_options=None,
+) -> Dict:
+    """
+    Build a hierarchical data model from a set of scanned grib messages. The iterable input groups should
```
martindurant, line 369:
```suggestion
    Build a hierarchical data model from a set of scanned grib messages. 
    
    The iterable input groups should
```

kerchunk/grib2.py
```diff
@@ -354,3 +359,197 @@ def example_combine(
...
+    :param message_groups: a collection of zarr store like dictionaries as produced by scan_grib
+    :param remote_options: remote options to pass to ZarrToMultiZarr
+    :return: A new zarr store like dictionary for use as a reference filesystem mapper
+    """
+    from kerchunk.combine import MultiZarrToZarr
+
+    # Hard code the filters in the correct order for the group hierarchy
+    filters = ["stepType", "typeOfLevel"]
+
+    zarr_store = {}
```
martindurant, line 388: Allow for parquet/bring-your-own output?

kerchunk/grib2.py
```diff
@@ -354,3 +359,197 @@ def example_combine(
...
+    zroot = zarr.open_group(store=zarr_store)
+    result = dict(refs=zarr_store)
+
+    aggregations: Dict[str, List] = defaultdict(list)
+    aggregation_dims: Dict[str, Set] = defaultdict(set)
+
+    unknown_counter = 0
+    for msg_ind, group in enumerate(message_groups):
+        if "version" not in result:
```
martindurant, line 397: Have you tested with zarr v3?

kerchunk/grib2.py
```diff
@@ -354,3 +359,197 @@ def example_combine(
...
+            raise RuntimeError(
+                f"Can not find a data var for msg# {msg_ind} in {group['refs'].keys()}"
+            )
+
+        if vname == "unknown":
+            # To resolve unknown variables add custom grib tables.
+            # https://confluence.ecmwf.int/display/UDOC/Creating+your+own+local+definitions+-+ecCodes+GRIB+FAQ
+            # If you process the groups from a single file in order, you can use the msg# to compare with the
+            # IDX file.
+            logger.warning(
```
martindurant, line 421: Why not include the message about what to do in the logging?

kerchunk/grib2.py
```diff
@@ -354,3 +359,197 @@ def example_combine(
...
+            if attr_val is None:
+                continue
+            if attr_val == "unknown":
+                logger.warning(
+                    "Found 'unknown' attribute value for key %s in var %s of msg# %s",
+                    key,
+                    vname,
+                    msg_ind,
+                )
+                # Use unknown as a group or drop it?
```
martindurant, line 442: Any cases like this?

kerchunk/grib2.py
martindurant: Actually, we need to figure out how parquet works at all for deep nested data like this.

martindurant (commented):
Do I understand that:
- each branch of the tree is combined as usual
- the branches have their keys copied into the right place in the overall output
- special renaming of grib keys is going on

kerchunk/grib2.py
```diff
@@ -354,3 +359,199 @@ def example_combine(
...
+    :param groups:
+    :param path:
+    :return:
+    """
+    result = []
+    for group in groups:
+        if "step/.zarray" not in group["refs"]:
+            group = copy.deepcopy(group)
+            logger.warning("Adding missing step variable to group path %s", path)
+            group["refs"]["step/.zarray"] = (
```
emfdavid, line 534: Can I put this in scan_grib where the cfgrib WrongStepUnitError is handled?
I don't like doing this here. Another alternative would be to use the preprocess hook in MultiZarrToZarr?
We should be able to calculate the correct step value by subtracting the valid_time from the time (reference time).
That requires decoding those values though and then encoding the step correctly...

emfdavid (commented):
Thanks for diving in - I hope this can be useful.

> * each branch of the tree is combined as usual

Yes - each variable can have different level values so only the branches can be combined in general

> * the branches have their keys copied into the right place in the overall output

Yes - similar to what you do in scan_grib, I am just adding a path for the zarr hierarchy.

> * special renaming of grib keys is going on

I did make copies of some attributes (name, stepType and typeOfLevel) to help users understand the hierarchy. I don't think I actually renamed any existing metadata. The implementation is opinionated about the grib level and grib step attributes as well as the coordinate names.

I would love to have more ECMWF data to test generality with.

kerchunk/grib2.py
emfdavid: In the HRRR SubHF dataset, even the step values that are decoded (don't raise an exception) are wrong.
```python
scans = scan_grib("gs://high-resolution-rapid-refresh/hrrr.20210928/conus/hrrr.t00z.wrfsubhf01.grib2")
dsets = []
zsets = []
vname = "dswrf"
for gg in scans:
  if f"{vname}/.zattrs" in gg['refs']:
    fs = fsspec.filesystem("reference", fo=gg)
    dsets.append(xr.open_dataset(fs.get_mapper(""), engine="zarr", consolidated=False))
    zsets.append(zarr.open_group(fs.get_mapper("")))
for ds, zs in zip(dsets, zsets):
  print(
      f"{ds[vname].attrs['GRIB_stepType']}, zs units {zs.step.attrs['units'] if 'step' in ds else None}: "
      f"ds step - {ds.step.values[()].astype('timedelta64[h]') if 'step' in ds else None}, zs step - {zs.step[()] if 'step' in zs else None};"
      f" valid_time - {ds.valid_time.values[()]}; time - {ds.time.values[()]}"
  )
```

```console
avg, zs units None: ds step - None, zs step - None; valid_time - 2021-09-28T00:15:00.000000000; time - 2021-09-28T00:00:00.000000000
instant, zs units hours: ds step - 15 hours, zs step - 15.0; valid_time - 2021-09-28T00:15:00.000000000; time - 2021-09-28T00:00:00.000000000
avg, zs units hours: ds step - 30 hours, zs step - 30.0; valid_time - 2021-09-28T00:30:00.000000000; time - 2021-09-28T00:00:00.000000000
instant, zs units hours: ds step - 30 hours, zs step - 30.0; valid_time - 2021-09-28T00:30:00.000000000; time - 2021-09-28T00:00:00.000000000
avg, zs units hours: ds step - 45 hours, zs step - 45.0; valid_time - 2021-09-28T00:45:00.000000000; time - 2021-09-28T00:00:00.000000000
instant, zs units hours: ds step - 45 hours, zs step - 45.0; valid_time - 2021-09-28T00:45:00.000000000; time - 2021-09-28T00:00:00.000000000
avg, zs units hours: ds step - 60 hours, zs step - 60.0; valid_time - 2021-09-28T01:00:00.000000000; time - 2021-09-28T00:00:00.000000000
instant, zs units hours: ds step - 1 hours, zs step - 1.0; valid_time - 2021-09-28T01:00:00.000000000; time - 2021-09-28T00:00:00.000000000
```

The values are in floating point minute but the units are specified as hours.
Only the stepType instant for the end of the forecast is correct - 1 hour.

emfdavid:
Not sure how to fix the Docs check but I think the rest of this is in pretty good shape!

TomNicholas:
@jthielen you might be interested in this

kerchunk/grib2.py
martindurant: Right, so version should always be 2? Or is this not the zarr version?

kerchunk/grib2.py
martindurant: OK, we can leave this for future improvement

kerchunk/grib2.py
martindurant: Are you saying that we should be able to decode them? The grib object has many many possble attributes, so this seems solvable. Not needed for this PR.

kerchunk/grib2.py
```diff
@@ -354,3 +360,216 @@ def example_combine(
...
+        for key, value in group["refs"].items():
+            if key not in [".zattrs", ".zgroup"]:
+                zarr_store[f"{path}/{key}"] = value
+
+    return result
+
+
+def correct_hrrr_subhf_step(group: Dict) -> Dict:
+    """
+    Overrides the definition of the step variable. Sets the value equal to the `valid_time - time`
```
martindurant, line 529:
```suggestion
    Overrides the definition of the "step" variable. 
    
    Sets the value equal to the `valid_time - time`
```

kerchunk/combine.py
```diff
@@ -311,8 +311,13 @@ def store_coords(self):
                         for _ in v
                     ]
                 ).ravel()
-            if "fill_value" not in kw and data.dtype.kind == "i":
-                kw["fill_value"] = None
+            if "fill_value" not in kw:
+                if data.dtype.kind == "i":
+                    kw["fill_value"] = None
+                elif k in z:
```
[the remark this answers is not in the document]
emfdavid: This is getting the `fill_value` attribute from the coordinate zarr array instance.
Without the if clause I got errors on some existing tests I think... so I copied the if clause used for copying the attributes (new line 329).
The change is important because timedeltas stored as float are given a default fill_value of 0.0 without this change which becomes NaT when read with xarray. Meaning you can't have a zero timedelta.

kerchunk/grib2.py
```diff
@@ -354,3 +360,218 @@ def example_combine(
...
+        a collection of zarr store like dictionaries as produced by scan_grib
+    remote_options: dict
+        remote options to pass to ZarrToMultiZarr
+
+    Returns
+    -------
+    list(dict): A new zarr store like dictionary for use as a reference filesystem mapper with zarr
+    or xarray datatree
+    """
+    from kerchunk.combine import MultiZarrToZarr
```
emfdavid, line 394: Can we move this to the top of the module?

kerchunk/grib2.py
emfdavid: Ah - you are right - this is the kerchunk version, not the zarr version anyway.

kerchunk/grib2.py
emfdavid: Crickets over at NOAA-EMC... I think we might have a working rust gribberish reader before we get a clear answer on how to build the ecCodes tables for these unknown properties.
🤞 @mpiannucci

kerchunk/grib2.py
emfdavid: Yes - I believe these are just more custom codes, similar to the unknown variables above.
I think they are all [documented](https://www.nco.ncep.noaa.gov/pmb/docs/grib2/grib2_doc/) and verifiable by reading the idx files, but implementing the tables for ecCodes to read them is more than I can take on right now...

emfdavid (commented):
@martindurant Thanks for taking a pass on this.
Do you want to add xarray_datatree as a dev dependency for kerchunk so we can make some additional tests?
Thank you for your work on datatree @TomNicholas any thoughts on how I am using or abusing datatree here?

kerchunk/grib2.py
mpiannucci: Yeah the tables are annoying, need to figure that out in rust too to be fully comprehensive

martindurant:
> Do you want to add xarray_datatree as a dev dependency for kerchunk so we can make some additional tests?

That's fine, add it to the CI envs. Having it in the optional package requirements is fine too, but not necessary (this is experimental, after all).

kerchunk/grib2.py
martindurant: OK

emfdavid:
> this is experimental, after all

Isn't it all experimental, after all?

martindurant:
> Isn't it all experimental, after all?

Some bits are more experimental than others :)

emfdavid:
@martindurant Still not sure what do about the docs build failure here?
https://github.com/fsspec/kerchunk/actions/runs/7050805133/job/19192272139?pr=399

martindurant:
> the docs build failure

I don't know either, except that the install raises a number of incompatibility warnings from pip. ammaraskar/sphinx-action@master doesn't appear to have changed, but it is somewhat dated - perhaps we need a concrete environment and run sphinx ourselves.

emfdavid:
Interesting... if I try to access the valid time coordinates from a non `square` aggregation e.g.
```python
sfcf_scans = scan_grib("testdata/hrrr.t00z.wrfsfcf00.grib2")
sfcf_scans += scan_grib("testdata/hrrr.t00z.wrfsfcf01.grib2")
sfcf_scans += scan_grib("testdata/hrrr.t01z.wrfsfcf00.grib2")
``` 
That is two forecast reference times, one of which as two steps and the other only has one step... we get some bad missing values in the valid time.

Screenshot 2023-11-30 at 4 30 01 PM

Where as if I combine outputs for two steps from each reference time...
```python
sfcf_scans = scan_grib("testdata/hrrr.t00z.wrfsfcf00.grib2")
sfcf_scans += scan_grib("testdata/hrrr.t00z.wrfsfcf01.grib2")
sfcf_scans += scan_grib("testdata/hrrr.t01z.wrfsfcf00.grib2")
sfcf_scans += scan_grib("testdata/hrrr.t01z.wrfsfcf01.grib2")
```

Screenshot 2023-11-30 at 4 39 08 PM

Maybe this is a behavior we could fix in MultiZarrToZarr to fill in NaN or NaT missing coordinate? I think the order in which you aggregate can also be important, but I am not sure why?

andreall:
Hi! I'm just getting started in the kerchunk and fsspec world and wanted to do fsspec on our grib files from WRF model. I was having a look at the colab notebook but I'm not sure how to implement it in my files...do I need to implement the correct_hrrr_subhf_step for my structure?

emfdavid:
> Is there some plugin point like that for datatree too?

Looks like [not yet](https://github.com/xarray-contrib/datatree/blob/66397e89529b2a3ecaa84d4945bd00da12a573a7/datatree/io.py#L37)?

martindurant:
I'm sure we could suggest it

emfdavid:
@martindurant pushed one more [commit](https://github.com/fsspec/kerchunk/pull/399/commits/229c3dae8db8297f9bb2fa4ad5146f399244e83b) 
Ran into issues using/testing the reference spec stores with byte values inside.
If there is a reason we need bytes I can roll this back, but I think you get back string anyway if you write/read it as json anyway?
If there is a way to fix this so the values are not created as bytes in the first place that would be ideal... it seems like the store data created by zarr with create_group and attr setter end up as binary consistent with the [zarr spec](https://zarr.readthedocs.io/en/stable/spec/v2.html#storage) but we end up with a mix of string/bytes in kerchunk - all string values, once you write/read it as json.
Screenshot 2023-12-01 at 12 40 50 PM

martindurant:
In JSON, the values are indeed strings on-disk, which become bytes when read by the filesystem: either as ascii, or base64 encoded. In memory (dicts), bytes, ascii strings or base64 strings should all work. Parquet, by contrast, stores bytes only.

keewis:
> I'm sure we could suggest it

See pydata/xarray#7437 (on the `xarray` repo because we've been thinking of merging `datatree` into `xarray` for quite a while now)

martindurant:
Merging this now, let's see what use it gets. We can always fix things for cases of gribs that are just a bit different from those trialed.

PR #404, opened 2023-12-15
Add MultiZarrToZarr.append()

martindurant:
cc @maxrjones

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -552,7 +552,7 @@ class MultiZarrToZarr:
         if filename is not None:
             with fsspec.open(filename, mode="wt", **(storage_options or {})) as f:
                 ujson.dump(out, f)
-        return out
+        return self.out
 
 
 def _reorganise(coos):
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -440,7 +440,7 @@ class MultiZarrToZarr:
     def second_pass(self):
         """map every input chunk to the output"""
         # TODO: this stage cannot be rerun without clearing and rerunning store_coords too,
-        #  because some code runs dependant on the current state of self.out
+        #  because some code runs dependent on the current state of self.out
         chunk_sizes = {}  #
         skip = set()
         dont_skip = set()
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -157,6 +157,38 @@ class MultiZarrToZarr:
         target_options=None,
         **kwargs,
     ):
+        """
+        Update an existing combined reference set with new references
+
+        There are two main usage patterns:
+
+        - if the input ``original_refs`` is JSON, the combine happens in memory and the
+          output should be written to JSON. This could then be optionally converted to parquet in a
+          separate step
+        - if ``original_refs`` is a lazy parquet reference set, then it will be amended in-place
+
+        If you want to extend JSON references and output to parquet, you must first convert to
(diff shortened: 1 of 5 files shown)


martindurant:
@rsignell-usgs , have you had a chance to do any experimenting with this?

rsignell:
@martindurant I've (finally) tested and both appending to JSON and Parquet are working fine!

https://nbviewer.org/gist/rsignell/ad1f3c66530a81e77585e05d9683984e

martindurant:
parquet part requires https://github.com/fsspec/filesystem_spec/pull/1510 (should be merged, then released soon)

kerchunk/combine.py
```diff
@@ -348,10 +440,11 @@ def store_coords(self):
     def second_pass(self):
         """map every input chunk to the output"""
         # TODO: this stage cannot be rerun without clearing and rerunning store_coords too,
-        #  because some code runs dependant on the current state f self.out
+        #  because some code runs dependant on the current state of self.out
```
rsignell, line 443: dependent

rsignell (approved):

kerchunk/combine.py
rsignell: @maxrjones do you have a few minutes to review?

martindurant: @maxrjones - particularly more tests would be good, although I'm happy to merge this early and add tests as we go

maxrjones: I'll have some time tomorrow afternoon to try it out

martindurant, line 443:
```suggestion
        #  because some code runs dependent on the current state of self.out
```

maxrjones: Sorry, I got caught up with other work but this is still on my list for this week

maxrjones:
@martindurant I added a Pythia chapter on the new feature in https://github.com/ProjectPythia/kerchunk-cookbook/pull/52. It works great, but I got stuck for a while on the need for the `coo_map={"time": "cf:time"}`, since the original MultiZarrtoZarr worked fine without specifying that. Am I missing something regarding how the coordinate alignment works usually?

martindurant:
> need for the coo_map={"time": "cf:time"}

I can't tell you how annoying these time units are in general for someone not from the field. I used to care about leap seconds  https://en.wikipedia.org/wiki/Terrestrial_Time , and barycentring, but it still seemed simpler.

> the original MultiZarrtoZarr worked fine without specifying that

It ... depends. Since the normal mode is to copy the values and attributes as-is, this will be no different to "cf:" and output dtype unspecified, if the inputs happened to have the same time epoch.

rsignell:
So just to make sure I've got it right:   If all the files have the same time units (e.g. "seconds since 1970-01-01 00:00") then we don't need it, but if the files in the collection we are combining have varying time units, like ["hours since 2012-01-01 00:00", "hours since 2013-01-01 00:00"] then we need the map.     Correct?

martindurant:
Correct, but having cf: in the first case shouldn't hurt either (except that you need cftime in the env, which is a given anyway for this case)

maxrjones:
hmm I think in https://github.com/ProjectPythia/kerchunk-cookbook/pull/52 all the files had the same time units, and excluding `cf:` worked in the original concatenation but failed when using append. I created a small test for demonstration in https://github.com/fsspec/kerchunk/pull/414, @martindurant should that be expected to work?

PR #414, opened 2024-01-30
Add test for appending with cftime

maxrjones:


martindurant:
This is probably a bad choice of unit
```
"units": "seconds since 1970-01-01T00:00:00"
```
because that's also the standard UNIX definition, so having no units would have the same effect.

A nice concise error:
```
numpy.core._exceptions._UFuncNoLoopError: ufunc 'less' did not contain a loop with signature matching types (<class 'numpy.dtypes.Int64DType'>, <class 'numpy.dtypes.DateTime64DType'>) -> None
```

I'll look into the error, though, I would have expected this to work.

martindurant:
OK, so of course xarray converts times automatically unless told otherwise.

maxrjones:
Thanks for the fix @martindurant! Did you not like the `drop_vars` option in the xarray backend?

Also, a bit unrelated but do you mind sharing with me what `...` passed to `__getitem__` does for zarr (referencing the following)?
```
o = z[selector.split(":", 1)[1]][...]
```

martindurant:
> Did you not like the drop_vars option in the xarray backend?

It was not being applied after the dataset was already loaded. This does the right thing for a normal load, but if some array fails to load (as I found with some time conversion), the options needs to be in the first open_dataset call. So you can use the argument as before, but it should work better now :)

> what ... passed to __getitem__ does

It would be the same as `[:]` (i.e., grab the actual values as an array) but it also works for zero-dimension arrays.

maxrjones:
The kerchunk tests take a long time to run in the CI, and it looks like most of the time is spent on the `setup-micromamba` step. Would you be open to enabling caching in setup-micromamba to speed up the workflows?

martindurant:
Certainly would be happy to have caching. Actually, conda now uses the mamba solver, so it might be faster again - I don't mind, so long as it works!

martindurant:
@maxrjones , env building on py39 seems stuck, any idea why?

maxrjones:
> @maxrjones , env building on py39 seems stuck, any idea why?

I pushed a couple fixes for the env errors

PR #417, opened 2024-02-08
Add parquet reference docs

martindurant:


martindurant:
This can be refined, but I'll merge it to make sure something is in the docs. 

The docstrings over at fsspec should ideally point here, and we should also make more tutorials on how to actually use these lazy things.

Issue #425, opened 2024-02-21
numcodecs version dependency


mullenkamp:
Hi,

I was wondering why you've locked the numcodecs version to <= 0.11 .
They are up to 0.12.1 now.

Thanks.

martindurant:
There was a thing with the entrypoints reading in numcodecs, but you are right we should keep up with their releases. You could make a PR relaxing the constraint and see if anything still breaks.

PR #426, opened 2024-02-21
relaxed numcodecs dependency

mullenkamp:


mullenkamp:
This relates to the issue I raised:
#425 
Seems good to align it to zarr dependencies.

martindurant:
I have confirmed locally that numcodecs 0.12.1 can see kerchunk's codecs without first importing kerchunk.

PR #439, opened 2024-03-21
add consolidate to single_zarr

raybellwaves:


kerchunk/zarr.py
```diff
@@ -1,11 +1,16 @@
...
 
 
 def single_zarr(
-    uri_or_store, storage_options=None, inline_threshold=100, inline=None, out=None
+    uri_or_store,
+    storage_options=None,
+    inline_threshold=100,
+    inline=None,
+    out=None,
+    translate=False,
```
martindurant: I think it should just always be True, no need for the kwarg. (Also "translate" is the method on most scanning classes, so would be confusing).
A good argument to optionally include would be an output; at the moment we have out is None (return JSON), or a parquet set; if it were a .json path, we should write to it.

raybellwaves: Apologies pushed an old commit. renamed it to consolidate

raybellwaves: I'll take a look at the write to json

raybellwaves:
I think this is ready for review.

I opened https://github.com/fsspec/kerchunk/issues/441 to think about writing to json in the function a bit more

PR #450, opened 2024-04-11
Add step:int remapping for coords computation

### LINKED ISSUE #446: Invalid try/except block in `scan_grib`?
In the `kerchunk.grib2.scan_grib()` function, there is a loop when processing the coordinate variables for a message, and there is a `try`/`except` block in this loop:

https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/grib2.py#L258-L267

that appears to be intended to skip coordinate variables that produce errors when retrieving the coordinate information with `cfgrib` (the `m` object is a `cfgrib.cfmessage.CfMessage` object).  However, exceptions produced in this `try` block never get caught because `cfgrib` only ever raises `ValueError`  or `AssertionError` exceptions (at least with the recent versions of `cfgrib`).

Should this `try` block be modified to catch `ValueError` exceptions instead?

kmpaul:
This is meant to address #446.

closes #446


dcherian:
(hi kevin!) 

Nice. Does this fix 

https://github.com/fsspec/kerchunk/blob/a0c4f3b828d37f6d07995925b324595af68c4a19/kerchunk/tests/test_grib.py#L78-L83

kmpaul:
(Hi, Deepak! :smile:)

That I do not know!  That is a very weird bug!  Which value for `step` is correct, `cfgrib` or `kerchunk`?

kmpaul:
@dcherian:  Sadly, it does not appear to fix the problem.  Sorry!

dcherian:
Just catch both and log whichever exception was caught? `eccodes` is a real pain, so I think most users would appreciate kerchunk being flexible with regard to eccodes version

kmpaul:
I've also verified that with the latest version of `eccodes`, the `kerchunk` tests fail when they encounter subhourly steps.  So, this would make the tests a bit more robust.

kmpaul:
I added another `ci/environment-py*.yml` file that explicitly installs the most recent version of `eccodes` for testing.  I can back this out, but I'm curious if these tests pass on CI.

kmpaul:
Incidentally, the reason `eccodes` is not being installed with the most recent version in the CI environment is because of the `h5py<3.9` dependency.  Can this restriction be removed (in one of the environment files, for example) for testing purposes?

kmpaul:
I may not be able to work on this any more today.  I might have to return to this tomorrow.

martindurant:
> I might have to return to this tomorrow.

No worries

> h5py<3.9 dependency. Can this restriction be removed

Let's try! I don't recall where this comes from.

kmpaul:
I've decided to deal with updating the tests in a separate PR so as not to entangle those changes with the changes specific to this PR.  I'll get back to this in a moment.

kmpaul:
Things have gotten a bit more complicated.  Let me explain:

1. With the latest versions of eccodes (2.34+), cfgrib no longer even produces `step` or `valid_time` in the resulting Xarray Dataset for subhourly data, so the `test_grib.py::test_archives[s3://noaa-hrrr-bdp-pds/hrrr.20140730/conus/hrrr.t23z.wrfsubhf08.grib2]` test fails because the `step` and `valid_time` coordinates are not in the Dataset for comparison in the test assert.  This will be fixed when cfgrib merges ecmwf/cfgrib#371, but for now I believe that kerchunk's `scan_grib` function is actually _more_ correct than cfgrib.  So, I am tempted to just skip this test for now.  But I am loathe to do that...

2. Also with the newest version of eccodes, the order of the variable attribute `"coordinates"` is not the same as it was before because (I presume) the `step` coordinate was added _ad hoc_ last.  But with the changes to eccodes and the change in this PR, the `step` coordinate doesn't have to be treated as "special", resulting in the `step" coordinate being added to the `"coordinate"` attribute in a new order.  This causes the assert in `test_grib.py::test_grib_tree` to fail, but it is easily fixed by changing the assert from a comparison of strings to a comparison of sets, where the sets are constructed from the `split()` of the strings.

3. The latest update to Xarray causes all time coordinates to be dtype `M8[ns]`, which causes the `test_combine.py::test_cftimes_to_normal` test to fail at the `dtype == "M8[s]"` assert.  To account to old and new versions of Xarray, I've changed the assert to `dtype.kind == "M"`, per @martindurant's recommendation.  But in the future this should probably be made more firm, and the `coo_dtypes` may need to be revisited if it doesn't do as expected because of these changes to Xarray.

kmpaul:
UPDATE:  I have marked the `test_grib.py::test_archives[s3://noaa-hrrr-bdp-pds/hrrr.20140730/conus/hrrr.t23z.wrfsubhf08.grib2]` test to be skipped if the version of eccodes >= 2.34.  This may need to be revisited.

kmpaul:
> Thanks for all the work here, hoping for greens

Me, too!

martindurant:
Didn't we have the h5py versions labeled as separate runs?

kmpaul:
I didn't make them separate.  I just removed the restriction in the `py310` and `py311` environment files.  Do you want me to make a separate environment?

kmpaul:
I didn't want to add too many test runs for fear of using too many resources, but I can add whatever environments you think are best.

martindurant:
(might be a good time to revert micromamba->miniconda, now that libmamba is a thing, perhaps that helps)

kmpaul:
Hey, they all passed!

kmpaul:
Do you still want me to make any further changes?

martindurant:
Oh, well that's good. 20min is still too long for 3.9, though!
I'll merge this, unblock the other PR, and we may choose to come back and add more pins to the env to make the build faster.

PR #458, opened 2024-06-04
made documentation updates

Anu-Ra-g:


docs/source/contributing.rst
```diff
@@ -124,7 +124,7 @@ Setting up pre-commit
 ~~~~~~~~~~~~~~~~~~~~~
 
 We use `pre-commit <https://pre-commit.com/>`_ to manage code linting and style.
-To set up pre-commit after activating your conda environment, run:
+To set up pre-commit **after activating** your conda environment, run:
```
martindurant, line 127: This is suggested, but not actually required - pre-commit established a git hook which will run in any environment. The only real precondition is that `pre-commit` (the command) is installed, which is easily done via conda or many other methods.

docs/source/contributing.rst
```diff
@@ -35,7 +35,7 @@ This project uses git for version control and github for issue tracking. If you
 Creating a Fork
 ---------------
 
-Once you have your git credentials setup, the next step is to create a fork of the project to work off of.
+Once you have your git credentials setup using this `tutorial <https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-ssh-urls>`_, the next step is to create a fork of the project to work off of.
```
martindurant, line 38: This doesn't look like the best link - you don't need SSH at all for this. I think the sentence essentially means "have an account on github", i.e., to have signed up.

Anu-Ra-g: Since the documentation didn't specify the https url for cloning the repo, I thought of adding this reference.

martindurant: That should then go with a separate instruction to clone the repo - which is a couple of lines below.

PR #460, opened 2024-06-07
add section to docs about the new dictionary feature

bendichter:


bendichter:
I think the phrasing could be better, but hopefully the example makes it clear

fyi @rly

docs/source/spec.rst
```diff
@@ -33,6 +33,18 @@ For example, Zarr data in this proposed spec might be represented as:
       "x/0": ["s3://bucket/path/file.nc", 294094376, 73825960]
     }
 
+Data can also be written in JSON, in which case the value is interpreted as a JSON file. For example, the above could equivalently be written as:
```
rly, line 36:
```suggestion
Data can also be written as a JSON object instead of a string, in which case the value is interpreted as a JSON file. For example, the above could equivalently be written as:
```

PR #466, opened 2024-06-21
Explicitly calculate dtype element size in netCDF3 records

### LINKED ISSUE #465: NetCDF file has one time step, kerchunk-generated reference has nine time steps?
This is pretty strange -- I have a 64-bit offset NetCDF file locally (because it doesn't work to open these from S3) and when I open it in xarray, I see one time step.  

If I generate the references for this file, somehow I'm getting 9 time steps.     Any ideas?

This notebook should be reproducible -- the NetCDF files are on AWS Open Data:
https://gist.github.com/rsignell/1ec34cac9d4583bb44cbddcace2b8657

martindurant:
Fixes #465 

@rsignell-usgs please test. Not what I thought was happening. I don't know why I can't trust numpy's `dt.itemsize`.

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -412,7 +412,6 @@ class MultiZarrToZarr:
                 elif k in z:
                     # Fall back to existing fill value
                     kw["fill_value"] = z[k].fill_value
-
             arr = group.create_dataset(
                 name=k,
                 data=data,
--- a/kerchunk/netCDF3.py
+++ b/kerchunk/netCDF3.py
@@ -285,7 +285,11 @@ class NetCDF3ToZarr(netcdf_file):
             }
         )
         if self.threshold:
-            out = inline_array(out, self.threshold, remote_options=self.storage_options)
+            out = inline_array(
+                out,
+                self.threshold,
+                remote_options=dict(remote_options=self.storage_options),
+            )
 
         if isinstance(out, LazyReferenceMapper):
             out.flush()
--- a/kerchunk/netCDF3.py
+++ b/kerchunk/netCDF3.py
@@ -1,4 +1,3 @@
-import base64
 from functools import reduce
 from operator import mul
 
@@ -6,7 +5,7 @@ import numpy as np
 from fsspec.implementations.reference import LazyReferenceMapper
 import fsspec
 
-from kerchunk.utils import _encode_for_JSON
+from kerchunk.utils import _encode_for_JSON, inline_array
 
 try:
     from scipy.io._netcdf import ZERO, NC_VARIABLE, netcdf_file, netcdf_variable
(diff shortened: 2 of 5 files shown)


rsignell:
@martindurant I tried testing, but don't know whether it's not working or simply user error opening in xarray (cell [8] in https://gist.github.com/rsignell/cb6e3ed842abedb797e2cd8ccc39169c

martindurant:
Should have been:
```
so = dict(anon=True, skip_instance_cache=True)
```
(remove `mode=`). Produces
```
<xarray.Dataset>
Dimensions:            (depth: 40, lat: 3251, lon: 4500, time: 1)
Coordinates:
  * depth              (depth) float64 0.0 2.0 4.0 6.0 ... 3e+03 4e+03 5e+03
  * lat                (lat) float64 -80.0 -79.96 -79.92 ... 89.92 89.96 90.0
  * lon                (lon) float64 -180.0 -179.9 -179.8 ... 179.8 179.8 179.9
  * time               (time) datetime64[ns] 2014-01-01T12:00:00
Data variables:
    salinity           (time, depth, lat, lon) float32 dask.array<chunksize=(1, 40, 3251, 4500), meta=np.ndarray>
    salinity_bottom    (time, lat, lon) float32 dask.array<chunksize=(1, 3251, 4500), meta=np.ndarray>
    surf_el            (time, lat, lon) float32 dask.array<chunksize=(1, 3251, 4500), meta=np.ndarray>
    water_temp         (time, depth, lat, lon) float32 dask.array<chunksize=(1, 40, 3251, 4500), meta=np.ndarray>
    water_temp_bottom  (time, lat, lon) float32 dask.array<chunksize=(1, 3251, 4500), meta=np.ndarray>
    water_u            (time, depth, lat, lon) float32 dask.array<chunksize=(1, 40, 3251, 4500), meta=np.ndarray>
    water_u_bottom     (time, lat, lon) float32 dask.array<chunksize=(1, 3251, 4500), meta=np.ndarray>
    water_v            (time, depth, lat, lon) float32 dask.array<chunksize=(1, 40, 3251, 4500), meta=np.ndarray>
    water_v_bottom     (time, lat, lon) float32 dask.array<chunksize=(1, 3251, 4500), meta=np.ndarray>
Attributes:
    Conventions:               CF-1.6 NAVO_netcdf_v1.1
    classification_authority:  not applicable
    classification_level:      UNCLASSIFIED
    distribution_statement:    Approved for public release. Distribution unli...
    downgrade_date:            not applicable
    field_type:                instantaneous
    history:                   archv2ncdf3z
    institution:               Naval Oceanographic Office
    source:                    HYCOM archive file
```

Note from our discussion: you can split any of the big arrays on the first dimension, which here is depth, length 40 (`kerchunk.utils.subchunk`). So long as you choose an exact divisor, e.g., 8 or 10, it will work. The on disk size unsplit is 1.17e9bytes, so probably worth doing this. I'm not actually sure if the split would be possible after comine, better to do on the simple JSON/dict from this first stage. (I **hope** I have counted for subchunking where the very first dimension is a 1, and so can be ignored.

martindurant:
Test should be fixed by https://github.com/fsspec/filesystem_spec/pull/1634

martindurant:
I'll merge this now and we can open any further issues that might arise.

rsignell:
Wooho!  It's working!

image

martindurant:
We'll need to decide what the best size is in the end. Having smaller chunks means more requests and a bigger reference set on disk. Since it's pretty easy to generate, we could make multiple variants and profile them.

PR #470, opened 2024-06-27
updated the ci envs

Anu-Ra-g:
Added support for `gcsfs` for the environments


Anu-Ra-g:
It is for this PR #471

martindurant:
I don't think we need this - I commented in the other PR. If you actually make a test that uses gcsfs, we shall see.

PR #471, opened 2024-06-27
added parse_grib_idx function

Anu-Ra-g:
Added `parse_grib_idx` function for parsing the `.idx` file in the cloud archives

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -589,14 +589,13 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
 def parse_grib_idx(
     basename: str,
     suffix: str = "idx",
-    storage_options: Dict = {},
-    tstamp: Optional["pd.Timestamp"] = None,
+    storage_options: Optional[Dict] = None,
     validate: bool = False,
 ) -> "pd.DataFrame":
     """
-    Standalone method used to extract metadata from a grib2 idx file(text) from NODD.
+    Parses per-message metadata from a grib2.idx file (text-type) to a dataframe of attributes
 
-    The function takes idx file, extracts the metadata known as attrs (variables with
+    The function uses the idx file, extracts the metadata known as attrs (variables with
     level and forecast time) from each idx entry and converts it into pandas
     DataFrame. The dataframe is later to build the one-to-one mapping to the grib file metadata.
 
@@ -608,8 +607,6 @@ def parse_grib_idx(
         The suffix is the ending for the idx file.
     storage_options: dict
         For accessing the data, passed to filesystem
-    tstamp : Optional[pd.Timestamp]
-        The timestamp to use for when the data was indexed
     validate : bool
         The validation if the metadata table has duplicate attrs.
 
@@ -619,20 +616,21 @@ def parse_grib_idx(
     """
     import pandas as pd
 
-    fs, _ = fsspec.core.url_to_fs(basename, **storage_options)
+    fs, _ = fsspec.core.url_to_fs(basename, **(storage_options or {}))
 
     fname = f"{basename}.{suffix}"
 
     baseinfo = fs.info(basename)
 
-    result = pd.read_csv(fs.open(fname), header=None, names=["raw_data"])
-    result[["idx", "offset", "date", "attrs"]] = result["raw_data"].str.split(
-        ":", expand=True, n=3
-    )
-    result["offset"] = result["offset"].astype(int)
+    with fs.open(fname) as f:
+        result = pd.read_csv(f, header=None, names=["raw_data"])
+        result[["idx", "offset", "date", "attrs"]] = result["raw_data"].str.split(
+            ":", expand=True, n=3
+        )
+        result["offset"] = result["offset"].astype(int)
 
-    # dropping the original single "raw_data" column before the formatting
-    result.drop(columns=["raw_data"], inplace=True)
+        # dropping the original single "raw_data" column after formatting
+        result.drop(columns=["raw_data"], inplace=True)
 
     result = result.assign(
         length=(
@@ -640,7 +638,6 @@ def parse_grib_idx(
         ),
         idx_uri=fname,
         grib_uri=basename,
-        indexed_at=tstamp if tstamp else pd.Timestamp.now(),
     )
 
     if validate and not result["attrs"].is_unique:
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -3,9 +3,11 @@ import copy
 import io
 import logging
 from collections import defaultdict
-from typing import Iterable, List, Dict, Set
-
+from typing import Iterable, List, Dict, Set, TYPE_CHECKING, Optional
 import ujson
+
+if TYPE_CHECKING:
+    import pandas as pd
 
 try:
     import cfgrib
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
     group["refs"]["step/0"] = enocded_val
 
     return group
+
+
+def parse_grib_idx(
+    fs: fsspec.AbstractFileSystem,
+    *,
+    basename: str,
+    suffix: str = "idx",
+    tstamp: Optional["pd.Timestamp"] = None,
+    validate: bool = False,
+) -> "pd.DataFrame":
+    """
+    Standalone method used to extract metadata from a grib2 idx file(text) from NODD.
+
+    The function takes idx file, extracts the metadata known as attrs (variables with
+    level and forecast time) from each idx entry and converts it into pandas
+    DataFrame. The dataframe is later to build the one-to-one mapping to the grib file metadata.
+
+    Parameters
+    ----------
+    fs : fsspec.AbstractFileSystem
+        The file system to read from.
+    basename : str
+        The base name is the full path to the grib file.
+    suffix : str
+        The suffix is the ending for the idx file.
+    tstamp : Optional[pd.Timestamp]
+        The timestamp to use for when the data was indexed
+    validate : bool
+        The validation if the metadata table has duplicate attrs.
+
+    Returns
+    -------
+    pandas.DataFrame : The data frame containing the results.
+    """
+    import pandas as pd
+
+    fname = f"{basename}.{suffix}"
+
+    baseinfo = fs.info(basename)
+
+    result = None
+
+    try:
+        result = pd.read_csv(fname, sep=":", header=None).loc[:, :5]
+        result.columns = ["idx", "offset", "date", "attrs", "level", "forecast"]
+        result["attrs"] = (
+            result["attrs"] + ":" + result["level"] + ":" + result["forecast"]
+        )
+        result.drop(columns=["level", "forecast"], inplace=True)
+    except Exception as e:
+        raise ValueError(f"Could not parse {fname}") from e
+
+    result = result.assign(
+        length=(
+            result.offset.shift(periods=-1, fill_value=baseinfo["size"]) - result.offset
+        ),
+        idx_uri=fname,
+        grib_uri=basename,
+        indexed_at=tstamp if tstamp else pd.Timestamp.now(),
+    )
+
+    if validate and not result["attrs"].is_unique:
+        raise ValueError(f"Attribute mapping for grib file {basename} is not unique)")
+
+    return result.set_index("idx")
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -587,10 +587,9 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
 
 
 def parse_grib_idx(
-    fs: fsspec.AbstractFileSystem,
-    *,
     basename: str,
     suffix: str = "idx",
+    storage_options: Dict = {},
     tstamp: Optional["pd.Timestamp"] = None,
     validate: bool = False,
 ) -> "pd.DataFrame":
@@ -603,12 +602,12 @@ def parse_grib_idx(
 
     Parameters
     ----------
-    fs : fsspec.AbstractFileSystem
-        The file system to read from.
     basename : str
         The base name is the full path to the grib file.
     suffix : str
         The suffix is the ending for the idx file.
+    storage_options: dict
+        For accessing the data, passed to filesystem
     tstamp : Optional[pd.Timestamp]
         The timestamp to use for when the data was indexed
     validate : bool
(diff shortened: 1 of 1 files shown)


kerchunk/grib2.py
```diff
@@ -582,3 +586,87 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+        result.offset.shift(periods=-1, fill_value=baseinfo["size"]) - result.offset
+    )
+
+    result.loc[:, "idx_uri"] = fname
+    result.loc[:, "grib_uri"] = basename
+    if tstamp is None:
+        tstamp = pd.Timestamp.now()
+    result.loc[:, "indexed_at"] = tstamp
+
+    if isinstance(fs, gcsfs.GCSFileSystem):
```
martindurant, line 641: Rather than having to import gcsfs (or s3fs), I recommend checking the fs.protocol or the existence of given fields in the info dict.

There is also an `fs.ukey()` method, which does what I think these blocks are intended for: test whether the remote file has changed or not. If you use this instead, you don't need the blocks at all. `.ukey` works for local files too.

Anu-Ra-g: Why do I have to check whether the remote files are changed or not?

Anu-Ra-g: But the code is just parsing the `.idx` files.

martindurant: Correct, so all the more reason not to require gcsfs

Anu-Ra-g: What other filesystems should I cover?

martindurant: If you use `ukey`, you automatically cover them all. 

Question: why do we need these keys at all?

kerchunk/grib2.py
```diff
@@ -3,7 +3,8 @@
 import io
 import logging
 from collections import defaultdict
-from typing import Iterable, List, Dict, Set
+from typing import Iterable, List, Dict, Set, Optional
+import pandas as pd
```
martindurant, line 7: Does kerchunk depend on pandas?
If this is only for the sake of typing, guard with
```
if typing.TYPE_CHECKING:
```

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+
+
+def parse_grib_idx(
+    fs: fsspec.AbstractFileSystem,
+    basename: str,
+    suffix: str = "idx",
+    tstamp: Optional[pd.Timestamp] = None,
+    validate: bool = False,
+) -> pd.DataFrame:
+    """
```
martindurant, line 595: We use numpydoc style in this repo, please reformat

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    Standalone method used to extract metadata from a grib2 idx file(text) from NODD
+    :param fs: the file system to read from
+    :param basename: the base name is the full path to the grib file
+    :param suffix: the suffix is the ending for the idx file
+    :param tstamp: the timestamp to record for this index process
+    :return: the data frame containing the results
```
martindurant, line 601: validate?

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    """
+
+    fname = f"{basename}.{suffix}"
+    fs.invalidate_cache(fname)
```
martindurant, line 605: Suggest removing these. If a caller needs to clear caches, that's on them.

kerchunk/grib2.py
Anu-Ra-g: Yes. Why do we need these keys? I used the `ukey` function. It is only generating a random string

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    fs.invalidate_cache(basename)
+
+    baseinfo = fs.info(basename)
+
+    with fs.open(fname, "r") as f:
+        splits = []
+        for line in f.readlines():
```
martindurant, line 612: Protect against big files
```suggestion
        for line in f:
```

This surely won't matted for idx files, but it's better style.

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+
+def parse_grib_idx(
+    fs: fsspec.AbstractFileSystem,
+    basename: str,
+    suffix: str = "idx",
+    tstamp: Optional[pd.Timestamp] = None,
+    validate: bool = False,
+) -> pd.DataFrame:
+    """
+    Standalone method used to extract metadata from a grib2 idx file(text) from NODD
```
martindurant, line 596: No real idea of what this function is for, how it might get called and how the results get used.

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    fs.invalidate_cache(fname)
+    fs.invalidate_cache(basename)
+
+    baseinfo = fs.info(basename)
+
+    with fs.open(fname, "r") as f:
+        splits = []
+        for line in f.readlines():
+            try:
+                idx, offset, date, attrs = line.split(":", maxsplit=3)
```
martindurant, line 614: Can pandas scan this file in a one-liner with read_csv(separator=":") ?

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+                        int(offset),
+                        f"{date[2:6]}-{date[6:8]}-{date[8:10]}-{date[10:]}",
+                        attrs,
+                    ]
+                )
+            except ValueError:
+                # Wrap the ValueError in a new one that includes the bad line
+                # If building the mapping, pick a different forecast run where the idx file is not broken
+                # If indexing a forecast using the mapping, fall back to reading the grib file
+                raise ValueError(f"Could not parse line: {line}")
```
martindurant, line 627:
```suggestion
            except ValueError as e:
                raise ValueError(f"Could not parse line: {line}") from e
```

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+
+    result = pd.DataFrame(
+        data=splits,
+        columns=["idx", "offset", "date", "attrs"],
+    )
+
+    # Subtract the next offset to get the length using the filesize for the last value
+    result.loc[:, "length"] = (
```
martindurant, line 635: This and the following lines could be covered with a single call to pandas.DataFrame.assign

kerchunk/grib2.py
```diff
@@ -582,3 +583,94 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+        result.offset.shift(periods=-1, fill_value=baseinfo["size"]) - result.offset
+    )
+
+    result.loc[:, "idx_uri"] = fname
+    result.loc[:, "grib_uri"] = basename
+    if tstamp is None:
+        tstamp = pd.Timestamp.now()
+    result.loc[:, "indexed_at"] = tstamp
+
+    if fs.protocol[0] == "gs":
```
martindurant, line 645: Without a justification of why we need all these fields, I prefer fs.ukey, which is sufficient to know whether the original files have changed.

martindurant (commented):
I realise that this code was not originally yours @Anu-Ra-g . However, this is the chance to produce better style and decide what information we want to parse and why.

We should attempt to include tests which cover every new line added.

kerchunk/grib2.py
martindurant: > It is only generating a random string. 

Are you sure it's _random_ ?

Anu-Ra-g: No I'm not sure. But how can the `ukey` function can be used to check the Filesystem?

martindurant: > No I'm not sure.

```python
    def ukey(self, path):
        """Hash of file properties, to tell if it has changed"""
        return sha256(str(self.info(path)).encode()).hexdigest()
```

kerchunk/grib2.py
Anu-Ra-g: This is for the `parse_grib_idx` function to return a Dataframe from parsing the `.idx` file

martindurant: > This is for the parse_grib_idx function to return a Dataframe from parsing the .idx file

The import can be inside the function, so that the module remains importable for those that don't have pandas.

martindurant: > the attrs field is only used from the Dataframe.

It would be good to document and test what attrs should be produced, so that the next stage can run.

kerchunk/grib2.py
```diff
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+
+    baseinfo = fs.info(basename)
+
+    result = None
+
+    try:
+        result = pd.read_csv(fname, sep=":", header=None).loc[:, :5]
+        result.columns = ["idx", "offset", "date", "attrs", "level", "forecast"]
+        result["attrs"] = (
+            result["attrs"] + ":" + result["level"] + ":" + result["forecast"]
```
martindurant, line 633: Why store these values as a single formatted string?

kerchunk/grib2.py
```diff
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    import pandas as pd
+
+    fname = f"{basename}.{suffix}"
+
+    baseinfo = fs.info(basename)
+
+    result = None
+
+    try:
+        result = pd.read_csv(fname, sep=":", header=None).loc[:, :5]
```
martindurant, line 630: This will fail is `fs` is not the local filesystem. I recommend you use path/storage_options as input as other kerchunk functions usually do.

kerchunk/grib2.py
```diff
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    -------
+    pandas.DataFrame : The data frame containing the results.
+    """
+    import pandas as pd
+
+    fname = f"{basename}.{suffix}"
+
+    baseinfo = fs.info(basename)
+
+    result = None
```
martindurant, line 627: This does nothing

kerchunk/grib2.py
```diff
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+
+    try:
+        result = pd.read_csv(fname, sep=":", header=None).loc[:, :5]
+        result.columns = ["idx", "offset", "date", "attrs", "level", "forecast"]
+        result["attrs"] = (
+            result["attrs"] + ":" + result["level"] + ":" + result["forecast"]
+        )
+        result.drop(columns=["level", "forecast"], inplace=True)
+    except Exception as e:
```
martindurant, line 636: We could choose to not catch this exception, and then the user just sees whatever pandas error underlies it.

kerchunk/grib2.py
```diff
@@ -582,3 +584,68 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
     group["refs"]["step/0"] = enocded_val
 
     return group
+
+
+def parse_grib_idx(
+    fs: fsspec.AbstractFileSystem,
+    *,
```
martindurant, line 591: Do you know what `*` does here? Is it reasonable?

martindurant (commented):
Obviously needs tests; and without the next stage of the workflow, this function isn't too useful by itself.

kerchunk/grib2.py
Anu-Ra-g: in the original code, the variables are stored with level and forecast as a single value

martindurant: Is that a good idea?

kerchunk/grib2.py
Anu-Ra-g: It's for the keyword based parameters. I assumed to pass the arguments after `fs` as keyword parameters.

martindurant: but why? The filename is *required*, and now all calling code needs to spell out `basename=` on every call.

> Is it reasonable?

kerchunk/grib2.py
Anu-Ra-g: Yes, it was failing. I needed both the grib file and idx file locally for the function to work.

kerchunk/grib2.py
Anu-Ra-g: Ok, I'll remove it.

kerchunk/grib2.py
Anu-Ra-g: `pd.read_csv("gs://global-forecast-system/gfs.20230901/00/atmos/gfs.t00z.pgrb2.0p25.f006.idx", sep=":", header=None)`
This was working

kerchunk/grib2.py
Anu-Ra-g: I don't know. What should I do here?

kerchunk/grib2.py
martindurant: We want it to be general. With publicly free data, any URL might work here, but often extra storage_options are required

kerchunk/grib2.py
martindurant: So what information do you need, what does the next stage require?

Anu-Ra-g: As there can be multiple variables of same name in a grib file, the original code combines the `variable:level:forecast time` in a string so that it can be used for the mapping

martindurant: Please explain this. I see no way you can "use" the variable/level/forecast values without unpacking the string again.

If you are worried about deciding whether values are unique, you can use `df.groupby(columns).count()`.

Anu-Ra-g: In the original code, `idx, offset, date, attrs = line.split(":", maxsplit=3)` are applied on the idx entry example, `162:137415046:d=2024042918:HGT:level of free convection:15 hour fcst:`

martindurant: So you are parsing the fields, sticking them together again and then splitting them up once more later...

Anu-Ra-g: This was the previous code. The dataframes used in the code used a combined value known as `attrs`

martindurant: I am aware of that. I wish to add only good code to this repo, so we need to be purposeful about how we do that.

kerchunk/grib2.py
```diff
@@ -582,3 +584,66 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    """
+    import pandas as pd
+
+    fs, _ = fsspec.core.url_to_fs(basename, **storage_options)
+
+    fname = f"{basename}.{suffix}"
+
+    baseinfo = fs.info(basename)
+
+    result = pd.read_csv(fs.open(fname), header=None, names=["raw_data"])
```
martindurant, line 628:
```suggestion
    with fs.open(fname) as f:
        result = pd.read_csv(f, header=None, names=["raw_data"])
```

kerchunk/grib2.py
```diff
@@ -582,3 +584,66 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
     group["refs"]["step/0"] = enocded_val
 
     return group
+
+
+def parse_grib_idx(
+    basename: str,
+    suffix: str = "idx",
+    storage_options: Dict = {},
```
martindurant, line 592:
```suggestion
    storage_options: Optional[Dict] = None,
```
It is considered bad practice to allow mutable default arguments. You will require `**storage_options` -> `**(storage_options or {})` below to conpensate.

kerchunk/grib2.py
```diff
@@ -582,3 +584,66 @@ def correct_hrrr_subhf_step(group: Dict) -> Dict:
...
+    tstamp: Optional["pd.Timestamp"] = None,
+    validate: bool = False,
+) -> "pd.DataFrame":
+    """
+    Standalone method used to extract metadata from a grib2 idx file(text) from NODD.
```
martindurant, line 597:
```suggestion
    Parses per-message metadata from a grib2 .idx file (text-type) to a dataframe of attributes
```

martindurant (approved):
In the following PR, when we have the other function too, we can actually make a test.

PR #475, opened 2024-07-10
Tolerate Zarr V3 module path name change

ghidalgo3:
The `encode_fill_value` function was moved into a new module in the latest version of `zarr`, and this causes a problem for packages that depend on both `zarr==3.0.0` and `kerchunk`, such as VirtualiZarr. I don't think this is necessarily related to https://github.com/fsspec/kerchunk/pull/292 , but that PR is free to cherry-pick this commit because they probably should be a single pull request.


ghidalgo3:
Thank you @martindurant!

PR #477, opened 2024-07-12
added extract_datatree_chunk_index function

Anu-Ra-g:


martindurant:
At this point, it would be really nice to have a workflow as a test, showing how these functions are to be used together. Unit tests of the individual functions, which are quite complex, would also be appreciated.

Anu-Ra-g:
Ok, I'll add those.

PR #485, opened 2024-07-29
added make_test_grib_idx function and tests

Anu-Ra-g:


kerchunk/tests/test_grib.py
```diff
@@ -346,3 +347,45 @@ def test_parse_grib_idx_content(idx_url, storage_options):
...
+        ),
+        (
+            "s3://noaa-hrrr-bdp-pds/hrrr.20220804/conus/hrrr.t00z.wrfsfcf01.grib2",
+            True,
+            dict(anon=True),
+            "hrrr.t00z.wrfsfcf01.grib2.test-limit-10",
+        ),
+    ],
+)
+def test_make_test_grib_idx_files_local(basename, local, storage_options, grib_file):
```
Anu-Ra-g, line 369: @martindurant I need your help here. I can't figure how this test is failing. I've tried two way but failing on both.

martindurant:
For local files, I recommend you use full absolute paths to avoid confusion. Also, if you want to test how your function works (which looks like a fixture, by the way), you can first copy the file to local rather than have two filesystems to get confused between.

Anu-Ra-g:
I thought `make_test_grib_idx_files` function was part of `dynamic zarr store` module so I added it. Should we add this to the `kerchunk` repo?

martindurant:
I don't understand - this PR is already wanting to add that function to the kerchunk repo (this repo).

martindurant:
If it isn't clear, my original comment above was in response to 
>  I need your help here. I can't figure how this test is failing. I've tried two way but failing on both.

which request of yours since disappeared.

Anu-Ra-g:
No, I meant you said this function looks like a fixture, so I asked the question.

martindurant:
Yes, the function could be made into a function and included in this repo. It is only used for tests, right?

Anu-Ra-g:
Yes, it will be used for tests.

martindurant:
Will it be used ONLY for tests? If yes, you can try making it a `@pytest.fixture`. It's purely a style thing, it seems like a fixture to me.

Anu-Ra-g:
> Also, if you want to test how your function works (which looks like a fixture, by the way), you can first copy the file to local rather than have two filesystems to get confused between.
The original implementation copied subset of the grib file to the cloud. I modified it so the subset file could be locally or to the cloud (if you have access to it).

martindurant:
> so the subset file could be locally or to the cloud 

Right, but apparently this wasn't working as you expected. Thus, it might be simpler to first copy a given file to local, and then make its .idx local->local.

Anu-Ra-g:
I meant test wasn't working. The function work fine.

martindurant:
If the test fails, maybe you can't be certain that the function is working fine.

Anu-Ra-g:
No I meant the test which I wrote for the `make_grib_test_files` functions are failing.

Anu-Ra-g:
Also this function isn't used anywhere in the actual flow or in the tests. There was bunch of grib-idx pairs the in the `asascience-open/nextgen-dmac` repository made by David. I waiting for his reply about what to do with this function.

martindurant:
> Also this function isn't used anywhere in the actual flow or in the tests.

?? 
Then why are you wanting to add it to the repo at all?

Anu-Ra-g:
It was in the original repo, so I thought of adding it. I'll close this PR.

PR #495, opened 2024-08-27
Added a new documentation page for faster GRIB aggregations

Anu-Ra-g:
This PR adds a new page in the `kerchunk` documentation for faster reference consolidation for GRIB files.

### DIFF
--- a/docs/source/reference_aggregation.rst
+++ b/docs/source/reference_aggregation.rst
@@ -1,30 +1,54 @@
-Other Methods of Aggregations
+Aggregation special cases
 =============================
 
 As we have already seen in this `page <https://fsspec.github.io/kerchunk/test_example.html#multi-file-jsons>`_,
-that the main purpose of ``kerchunk`` it to generate references, to view whole archive of files like
-GRIB2, NetCDF etc. allowing us for *in-situ* access to the data. In this part of the documentation,
-we will see some other efficient ways of combining references.
+that the main purpose of ``kerchunk`` it to generate references, to view whole archive
+of files like GRIB2, NetCDF etc. allowing us for direct access to the data. In
+this part of the documentation, we will see some other efficient ways of
+combining references.
 
 GRIB Aggregations
 -----------------
 
-This new method for reference aggregation, developed by **Camus Energy**, is based on GRIB2 files. Utilizing
-this method can significantly reduce the time required to combine references, cutting it down to
-a fraction of the previous duration. In reality, this approach builds upon consolidating references
-with ``kerchunk.combine.MultiZarrtoZarr``, making it faster. The functions and operations used in this
-will be a part of the ``kerchunk``'s API. You can follow this `video <https://discourse.pangeo.io/t/pangeo-showcase-optimizations-for-kerchunk-aggregation-and-zarr-i-o-at-scale-for-machine-learning/4074>`_
-for the intial discussion.
+This reference aggregation method of GRIB files, developed by **Camus Energy**, functions if
+accompanying ``.idx`` files are present.
+
+**But this procedure has some certain restrictions:**
+
+  - GRIB files must paired with their ``.idx`` files
+  - The ``.idx`` file must be of *text* type.
+  - Only specialised for time-series data, where GRIB files
+    have *identical* structure.
+  - Aggregation only works for a specific **forecast horizon** files.
+
+Utilizing this method can significantly reduce the time required to combine
+references, cutting it down to a fraction of the previous duration. The original
+idea was showcased in this `talk <https://discourse.pangeo.io/t/pangeo-showcase-optimizations-for-kerchunk-aggregation-and-zarr-i-o-at-scale-for-machine-learning/4074>`_.
 
 *How is it faster*
 
-Every GRIB file stored on cloud platforms such as **AWS** and **GCP** is accompanied by its
-corresponding ``.idx`` file. This file otherwise known as an *index* file contains the key
+The ``.idx`` file otherwise known as an *index* file contains the key
 metadata of the messages in the GRIB files. These metadata include `index`, `offset`, `datetime`,
 `variable` and `forecast time` for their respective messages stored in the files.
 
+**It follows three step approach:**
+
+  1. Extract and persist metadata directly from a few arbitrary grib
+     files for a given product such as HRRR SUBH, GEFS, GFS etc.
+  2. Use the metadata mapping to build an index table of every grib
+     message from the ``.idx`` files
+  3. Combine the index data with the metadata to build any FMRC
+     slice (Horizon, RunTime, ValidTime, BestAvailable)
+
+.. tip::
+  To confirm the indexing of messages, see this `notebook <https://gist.github.com/Anu-Ra-g/efa01ad1c274c1bd1c14ee01666caa77>`_.
+
 These metadata will be used to build an k_index for every GRIB message that we will be
-indexing. The indexing process primarily involves the `pandas <https://pandas.pydata.org/>`_ library.
+indexing. Indexing process primarily involves the `pandas <https://pandas.pydata.org/>`_ library.
+
+.. note::
+    The index in ``.idx`` file indexes the GRIB messages where as the ``k_index`` (kerchunk index)
+    we build as part of this workflow index the variables in those messages.
 
 .. list-table:: k_index for a single GRIB file
    :header-rows: 1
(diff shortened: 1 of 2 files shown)


docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
+Other Methods of Aggregations
```
martindurant, line 1: I would cal it
"Aggregation special cases"

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+=============================
+
+As we have already seen in this `page <https://fsspec.github.io/kerchunk/test_example.html#multi-file-jsons>`_,
+that the main purpose of ``kerchunk`` it to generate references, to view whole archive of files like
+GRIB2, NetCDF etc. allowing us for *in-situ* access to the data. In this part of the documentation,
```
martindurant, line 6: in-situ -> direct
(I think in-situ here would imply that the processing is happening wherever the data is, which is not the case; I know what you mean is that we read the originals, not a copy)

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+we will see some other efficient ways of combining references.
+
+GRIB Aggregations
+-----------------
+
+This new method for reference aggregation, developed by **Camus Energy**, is based on GRIB2 files. Utilizing
```
martindurant, line 12: I won't be "new" for long, so drop this.

I would put the restrictions first:
- must have .idx files
- specialised for time-series data, each file having identical message structure

@emfdavid : an opinion on whether Camus wants to be referenced here and how that would look.

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+this method can significantly reduce the time required to combine references, cutting it down to
+a fraction of the previous duration. In reality, this approach builds upon consolidating references
+with ``kerchunk.combine.MultiZarrtoZarr``, making it faster. The functions and operations used in this
+will be a part of the ``kerchunk``'s API. You can follow this `video <https://discourse.pangeo.io/t/pangeo-showcase-optimizations-for-kerchunk-aggregation-and-zarr-i-o-at-scale-for-machine-learning/4074>`_
+for the intial discussion.
```
martindurant, line 17:
```suggestion
a fraction of the previous duration. The original idea was showcased in this talk <link>
```

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+
+*How is it faster*
+
+Every GRIB file stored on cloud platforms such as **AWS** and **GCP** is accompanied by its
```
martindurant, line 21: Not true! **IF** there are .idx files, we can use them.

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+corresponding ``.idx`` file. This file otherwise known as an *index* file contains the key
+metadata of the messages in the GRIB files. These metadata include `index`, `offset`, `datetime`,
+`variable` and `forecast time` for their respective messages stored in the files.
+
+These metadata will be used to build an k_index for every GRIB message that we will be
```
martindurant, line 26: What is a k_index?

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+     - 2017-01-01 12:00:00
+     - s3://noaa-gefs-pds/gefs.20170101/06/gec00.t06z...
+     - 100391
+     - 20488
+     - None
+
+
+.. note::
+    The index in ``idx`` file indexes the GRIB messages where as the ``k_index`` (kerchunk index)
+    we build as part of this workflow index the variables in those messages.
```
martindurant, line 115: This note explains my question above, but is not very clear. Map out the steps we need to do before launching into the details.

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,191 @@
...
+
+*What now*
+
+After creating the k_index as per the desired duration, we will use the ``DataTree`` model
+from the `xarray-datatree <https://xarray-datatree.readthedocs.io/en/latest/>`_ to view a
+part of the aggregation or the whole. Below is a tree model made from an aggregation of
```
martindurant, line 121: Did the process finish, is this the final aggregated whole?

Anu-Ra-g: I was using the original code for reinflating. So I mentioned it.

docs/source/reference_aggregation.rst
Anu-Ra-g: Steps for how we build the index? Should I include the code for building the index?

docs/source/reference_aggregation.rst
Anu-Ra-g: The grib and idx files existed in pairs, so I mentioned it.

docs/source/reference_aggregation.rst
```diff
@@ -1,30 +1,54 @@
...
+This reference aggregation method of GRIB files, developed by **Camus Energy**, functions if
+accompanying ``.idx`` files are present.
+
+**But this procedure has some certain restrictions:**
+
+  - GRIB files must paired with their ``.idx`` files
+  - The ``.idx`` file must be of *text* type.
+  - Only specialised for time-series data, where GRIB files
+    have *identical* structure.
+  - Aggregation only works for a specific **forecast horizon** files.
```
emfdavid, line 22: The reference index can be combined across many horizons but each horizon must be indexed separately.
Looking forward to seeing what you make of the reinflate api... there you can see all of the FMRC slices are supported agains a collection of indexed data from many horizons, runtimes and valid times.

emfdavid (commented):
Great start - one suggestion re limitations.

docs/source/reference_aggregation.rst
martindurant: > the reinflate api

ooh, what is this?

emfdavid: The method to turn the k_index and the metadata back into a ref_spec you can use in zarr/xarray
https://github.com/asascience-open/nextgen-dmac/blob/main/grib_index_aggregation/dynamic_zarr_store.py#L198
I think @Anu-Ra-g is already working on adding it into kerchunk?

martindurant: Oh, in that case I suspect it works already, right @Anu-Ra-g : but you can only work on one set of horizons OR one set of timepoints, not both at once? Something like that.

emfdavid: I think you can return an array with multiple dimensions.
I didn't have a strong use for this so struggled to do something general and practical.
For instance if you request by Horizon, you can provide multiple horizon axis and you dimensions should include 'horizon' and 'valid_time". Similarly you can request multiple runtimes and then your dimensions should include 'runtime' and 'step'.
Honestly not sure if this is helpful or over complicated.

Anu-Ra-g: @martindurant I tried it out with one set of horizons with the original code. Actually, I'm still figuring out the reinflating  part of the code,  aggregation types and the new indexes. 

I noticed that for reinflating can also work with a `grib_tree` model made from a **single grib** file. 
@emfdavid can you confirm this in this notebook that I made?

docs/source/reference_aggregation.rst
Anu-Ra-g: @emfdavid  Should I mention `Camus Energy` like @martindurant  asked here?

emfdavid: I checked in with the team and it would be great if you could link https://www.camus.energy/ and if there is a place for attribution you can include my github handle @emfdavid, otherwise the contributors list is fine.

docs/source/reference_aggregation.rst
martindurant: No code, just brief points.

martindurant:
Let me know when this PR is ready for another look.

Anu-Ra-g:
@martindurant  I've made the changes like you've suggested. It is ready for review.

docs/source/reference_aggregation.rst
```diff
@@ -0,0 +1,224 @@
...
+                        valid_time  (model_horizons, valid_times) datetime64[ns] 2kB ...
+                    Dimensions without coordinates: model_horizons, valid_times
+                    Data variables:
+                        ulwrf       (model_horizons, valid_times, latitude, longitude) float64 124MB ...
+                    Attributes:
+                        typeOfLevel:  nominalTop
+
+
+.. tip::
+    For a full tutorial on this workflow, refer this `kerchunk cookbook <https://projectpythia.org/kerchunk-cookbook/README.html>`_
```
martindurant, line 218: Should this link to the new notebook you are adding in https://github.com/ProjectPythia/kerchunk-cookbook/pull/64 ?

Anu-Ra-g: I was going to update it here the after the notebook in the pythia got merged.

PR #498, opened 2024-08-30
added support for s3 files

Anu-Ra-g:
This PR adds the `remote_options` parameter in necessary functions to be compatible with S3.

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -813,8 +813,7 @@ def extract_datatree_chunk_index(
 def _map_grib_file_by_group(
     fname: str,
     mapper: Optional[Callable] = None,
-    storage_options=None,
-    remote_options: Dict = None,
+    storage_options: Optional[Dict] = None,
 ) -> "pd.DataFrame":
     """
     Helper method used to read the cfgrib metadata associated with each message (group) in the grib file
@@ -844,7 +843,7 @@ def _map_grib_file_by_group(
                 filter(
                     lambda item: item is not None,
                     [
-                        _extract_single_group(mapper(group), i, remote_options)
+                        _extract_single_group(mapper(group), i, storage_options)
                         for i, group in enumerate(references, start=1)
                     ],
                 )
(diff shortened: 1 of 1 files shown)


Anu-Ra-g:
@martindurant It was in my local but I didn't push in the original mapping branch.

kerchunk/grib2.py
```diff
@@ -811,7 +811,10 @@ def extract_datatree_chunk_index(
 
 
 def _map_grib_file_by_group(
-    fname: str, mapper: Optional[Callable] = None, storage_options=None
+    fname: str,
+    mapper: Optional[Callable] = None,
+    storage_options=None,
+    remote_options: Dict = None,
```
martindurant, line 817: Why is storage_options and remote_options different? We are only reading one remote file, no?

Aside: the type annotation should be `Optional[dict]` or `dict | None`.

Anu-Ra-g: It is for the remote_options parameter in `grib_tree`
```
def grib_tree(
    message_groups: Iterable[Dict],
    remote_options=None,
) -> Dict:
```

martindurant: But why is it different

Anu-Ra-g: Does `remote_options` and `target_options` come under the `storage_options`?

martindurant: I don't understand your question. You are trying to add both storage_options and remote_options to this function signature, and I don't see why they would be different - both reference the same remote grib file(s), no?

Anu-Ra-g: I though it'd be confusing. So I made them different.  I'll make it `storage_options` across the code.

Anu-Ra-g: I got confused with the signature of `grib_tree` function,
```
def grib_tree(
    message_groups: Iterable[Dict],
    remote_options=None,
   ```

Anu-Ra-g: @martindurant  I've made the updates.

PR #499, opened 2024-08-30
added the reinflate api

Anu-Ra-g:
Added the `reinflate` function to reinflate the index.

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -1,12 +1,15 @@
 import base64
 import copy
 import io
+import os
 import logging
 from collections import defaultdict
 import warnings
-from typing import Iterable, List, Dict, Set, TYPE_CHECKING, Optional, Callable
+from enum import unique, Enum
+from typing import Iterable, List, Dict, Set, TYPE_CHECKING, Optional, Callable, Any
 import ujson
 import itertools
+import re
 
 if TYPE_CHECKING:
     import pandas as pd
@@ -20,6 +23,25 @@ import numpy as np
 from kerchunk.utils import class_factory, _encode_for_JSON
 from kerchunk.codecs import GRIBCodec
 from kerchunk.combine import MultiZarrToZarr, drop
+
+COORD_DIM_MAPPING: dict[str, str] = dict(
+    time="run_times",
+    valid_time="valid_times",
+    step="model_horizons",
+)
+
+
+@unique
+class AggregationType(Enum):
+    """
+    ENUM for aggregation types
+    TODO is this useful elsewhere?
+    """
+
+    HORIZON = "horizon"
+    VALID_TIME = "valid_time"
+    RUN_TIME = "run_time"
+    BEST_AVAILABLE = "best_available"
 
 
 try:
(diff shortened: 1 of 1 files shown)


kerchunk/grib2.py
```diff
@@ -975,3 +1135,257 @@ def build_idx_grib_mapping(
...
+
+
+def reinflate_grib_store(
+    axes: list["pd.Index"],
+    aggregation_type: AggregationType,
+    chunk_index: "pd.DataFrame",
+    zarr_ref_store: dict,
+) -> dict:
+    """
+    Given a zarr_store hierarchy, pull out the variables present in the
```
martindurant, line 1225: There should be a one-line summary at the start of the docstring, since this is what help() and sphynx use as the title describing the method.

martindurant (commented):
The functions here are identical to the originals by @emfdavid ?

I am OK to include these now, but I think the next PR should move all of the tree/idx-related functions to another module (_grib_idx ?), and only import the public-facing functions into this module. As it is, the module is now extremely long compared to other formats, and it's not so obvious where a user should start.

emfdavid:
Separate module and only exposing public functions sounds good to me.

emfdavid:
Is there a better name for this function? I am terrible at names...

martindurant:
> Is there a better name for this function?

You mean "reinflate"? Maybe if you write a 2-sentence description, the name will present itself :)

Anu-Ra-g:
@martindurant  what should be the module name? The module should be included in a sub-package like `kerchunk.grib.<module>` or directly in the `kerchunk.<module name>`?

martindurant:
I suggested `_grib_idx` above. Making `kerchunk.grib` into a package sounds like a _bad_ idea.

emfdavid:
Comments on naming (not my strong suit)... the Camus team had suggested "outbasing" as the name for this general approach of ingesting an index of chunks to a database and constructing the dataset/tree on the fly.

The "reinflate" method in particular builds the refspec dictionary for a zarr/xarray tree representing a set of grib data. The inputs are the static metadata for the refspec (everything but the chunks for variables with a time dimension), a set of chunk indexes for specific variables (usually stored in a DB), the aggregation type, and the set of axes that define how to arrange those chunks.

I wanted to be general, supporting all the [FMRC](https://docs.unidata.ucar.edu/netcdf-java/current/userguide/images/netcdf-java/tutorial/feature_types/fmrc.png) slices which are represented in the [AggregationType](https://github.com/fsspec/kerchunk/pull/499/files#diff-513711a9f717c9a6a924112e96101dcb96f8fe751f0c2590c65e54e8001c7501R41-R44) enum. The api allows passing a list of axes such that you can express a set of runtimes, a set of valid times, or a set of horizons.

I hope there is some simpler expression of these concepts that might use this method as a backend, but I suspect you would have to make assumptions about the backend database/schema to wrap it up nicely.

martindurant:
- This code should be in _grib_idx.py, as discussed
- There should be tests
- There needs to be far more documentation on how to use the code (again), else no one will ever touch these functions.

emfdavid:
Thank you for getting the process started. I will plan to make progress this month, but it will be a nights and weekends effort. If that is not progressing by the end of October we can regroup.

martindurant:
Sure, I was simply book-keeping :)

rsignell:
@emfdavid I was trying to figure out where the GSoC work stands, and it looks like this is the blocker?  Is that correct?

martindurant:
Check out #523 - I think that's intended to finish off the topic.

rsignell:
@martindurant ooh, yes indeed, that looks like it!  Thanks!

emfdavid:
Can we close this now that #523 is merged?

PR #500, opened 2024-09-08
added new module and necessary refactorings

Anu-Ra-g:
This PR adds a new module called `_grib_idx.py` with private functions and added necessary refactorings


martindurant:
How does this one relate to #499 ?

kerchunk/_grib_idx.py
```diff
@@ -0,0 +1,250 @@
+import ujson
```
martindurant, line 1: Since this module does not have user functions (which should be exposed via `kerchunk.grib`), I think the whole module ought to have a docstring to explain this.

kerchunk/grib2.py
```diff
@@ -655,236 +649,6 @@ def parse_grib_idx(
...
-        return None
-
-    assert (
-        len(k_ind) == 1
-    ), f"expected a single variable grib group but produced: {k_ind}"
-    k_ind.loc[:, "idx"] = idx
-    return k_ind
-
-
 def build_idx_grib_mapping(
```
martindurant, line 652: Why did this function not move? I would think it should live in `_grib_index`, but it can be _imported_ here, to indicate that, yes, it is user-facing.

martindurant (commented):
parse_grib_idx and other functions are definitely part of the idx workflow only, I expected they would be moved.

martindurant:
> the pre-commit is failing on the unused imports

You can fix this by including an `__all__`, which should be a list of strings to be considered as "public" in the module. Making this list has two effects:
- making the imports look like they are used
- defining what gets imported if you were to do `from kerchunk.grib import *`.

Anu-Ra-g:
Should I add the original functions like `scan_grib`, `grib_tree` etc to the list or only the indexing functions?
`__all__ = ["build_idx_grib_mapping", "map_from_index"]`

martindurant:
Yes please, add all the public functions, just incase some one does do a "*" import.

Anu-Ra-g:
@martindurant made the changes

PR #506, opened 2024-09-21
Allow disabling inlining for MultiZarrToZarr

maresb:
I am trying to combine some massive S3-based Zarr references. A `MultiZarrToZarr` operation which would otherwise take a minute would take days because the size of each file on S3 is being queried. Thus I want to be able to pass in `inline_threshold=None` so that my datasets combine in a reasonable timeframe.

### DIFF
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -94,7 +94,7 @@ class MultiZarrToZarr:
         If True, will load the references specified by out and add to them rather than starting
         from scratch. Assumes the same coordinates are being concatenated.
     """
-    
+
     inline: int
 
     def __init__(
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -1,7 +1,7 @@
 import collections.abc
 import logging
 import re
-from typing import List, Optional
+from typing import List
 import warnings
 
 import fsspec
@@ -78,9 +78,8 @@ class MultiZarrToZarr:
     :param remote_protocol: str
         The protocol of the original data
     :param remote_options: dict
-    :param inline_threshold: int | None
-        Size below which binary blocks are included directly in the output. If None, no
-        inlining is done.
+    :param inline_threshold: int
+        Size below which binary blocks are included directly in the output
     :param preprocess: callable
         Acts on the references dict of all inputs before processing. See ``drop()``
         for an example.
@@ -95,6 +94,8 @@ class MultiZarrToZarr:
         If True, will load the references specified by out and add to them rather than starting
         from scratch. Assumes the same coordinates are being concatenated.
     """
+    
+    inline: int
 
     def __init__(
         self,
@@ -107,7 +108,7 @@ class MultiZarrToZarr:
         target_options=None,
         remote_protocol=None,
         remote_options=None,
-        inline_threshold: Optional[int] = 500,
+        inline_threshold: int = 500,
         preprocess=None,
         postprocess=None,
         out=None,
@@ -586,7 +587,7 @@ class MultiZarrToZarr:
 
                     ref = fs.references.get(fn)
                     if (
-                        self.inline is not None
+                        self.inline > 0
                         and isinstance(ref, list)
                         and (
                             (len(ref) > 1 and ref[2] < self.inline)
--- a/kerchunk/combine.py
+++ b/kerchunk/combine.py
@@ -1,7 +1,7 @@
 import collections.abc
 import logging
 import re
-from typing import List
+from typing import List, Optional
 import warnings
 
 import fsspec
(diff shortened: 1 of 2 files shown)


martindurant:
I think this was already possible with inline_threashold=0 ?

maresb:
Nope, if `inline_threshold=0` then `fs.info(fn)` still gets called in order to ensure that the file size isn't negative. :sweat_smile:

We could change this so that we use `0` in place of `None`. That would probably make more sense actually. Then my conditional `self.inline is not None` becomes `self.inline > 0`.

martindurant:
That sounds like a good plan

maresb:
Ok @martindurant, that should take care of it.

Now the code should be functionally identical to the previous, except that the file size lookup is skipped when unneeded. In other words, no more `inline_threshold=None` but instead `inline_threshold=0` is optimized.

kerchunk/combine.py
```diff
@@ -94,6 +94,8 @@ class MultiZarrToZarr:
         If True, will load the references specified by out and add to them rather than starting
         from scratch. Assumes the same coordinates are being concatenated.
     """
+
```
martindurant, line 97:
```suggestion

```

kerchunk/combine.py
```diff
@@ -94,6 +94,8 @@ class MultiZarrToZarr:
         If True, will load the references specified by out and add to them rather than starting
         from scratch. Assumes the same coordinates are being concatenated.
     """
+    
+    inline: int
```
martindurant, line 98:
```suggestion

    inline: int
```

maresb:
Oh, I'm sorry, I'm on my laptop at the moment and it looks like I forgot to set up pre-commit :see_no_evil: 

Please feel free to squash this PR.

maresb:
Hmm, this test failure I can reproduce locally, also from `main`. I'm guessing it's a dependency issue.

martindurant:
@Anu-Ra-g , any idea?
```
>       assert (
            idx_df["typeOfLevel"][[42, 46, 49, 50]]
            == ["heightAboveGround", "isobaricInhPa", "surface", "heightAboveGround"]
        ).all()
E       AssertionError: assert False
E        +  where False = <bound method NDFrame._add_numeric_operations.<locals>.all of 42     True\n46     True\n49     True\n50    False\nName: typeOfLevel, dtype: bool>()
E        +    where <bound method NDFrame._add_numeric_operations.<locals>.all of 42     True\n46     True\n49     True\n50    False\nName: typeOfLevel, dtype: bool> = 42    heightA... dtype: object == ['heightAbove...tAboveGround']
E             Full diff:
E             - [
E             -  'heightAboveGround',
E             ? ^^                 --
E             + 42    heightAboveGround
E             ? ^^^^^^
E             -  'isobaricInhPa',...
E
E             ...Full output truncated (8 lines hidden), use '-vv' to show.all
```

maresb:
Interesting, it seems like something's getting mangled. When I run locally the assert above passes for me, but the subsequent one fails.

Note that for me `stepType` is `instant` instead of the asserted `accum`.

```python
idx_df.iloc[83]
varname                                                             wz
typeOfLevel                                              isobaricInhPa
stepType                                                       instant
name                                       Geometric vertical velocity
isobaricInhPa                                                    975.0
step                                                   0 days 06:00:00
time                                               2023-09-28 00:00:00
valid_time                                         2023-09-28 06:00:00
uri                  ~/repos/kerchunk/kerchunk/tests/gfs....
offset                                                        21234675
length                                                         1035696
inline_value                                                      None
surface                                                            NaN
heightAboveGround                                                  NaN
meanSea                                                            NaN
Name: 83, dtype: object
GitHub request
```

maresb:
> I'm guessing it's a dependency issue.

The culprit is [eccodes v2.38.0](https://github.com/ecmwf/eccodes/releases/tag/2.38.0).

Anu-Ra-g:
@maresb I tried to emulate the above steps on my local system. The error is not occurring consistently.
And this object is still `accum` when I tried it locally,
```
varname                                                           watr
typeOfLevel                                                    surface
stepType                                                         accum
name                                                      Water runoff
isobaricInhPa                                                      NaN
step                                                   0 days 06:00:00
time                                               2023-09-28 00:00:00
valid_time                                         2023-09-28 06:00:00
uri                  ./kerchunk/tests/gfs.t00z.pgrb2.0p25.f006.test...
offset                                                        59215098
length                                                          348767
inline_value                                                      None
surface                                                            0.0
heightAboveGround                                                  NaN
meanSea                                                            NaN
Name: 83, dtype: object
```

maresb:
Thanks @Anu-Ra-g! May I ask which version of eccodes you are using?

Anu-Ra-g:
I'm on eccodes version `2.36.0`

maresb:
Since the test failure is unrelated to this PR, I recommend moving discussion to #508.

@Anu-Ra-g, to reproduce the failure you need to upgrade to eccodes v2.38.0.

maresb:
Thanks a lot @martindurant for the merge!!! :pray:

PR #512, opened 2024-09-28
Added initial code for local index definition of grib files.

Anu-Ra-g:
This is thed demo for an initial version of the local grib index file defintions. The code still needs more update. Below is the output  created in a `.idx` file
```
1:0:d=2023092800:V:6 hour fcst
2:955132:d=2023092800:ABSV:6 hour fcst
3:1898414:d=2023092800:CLWMR:6 hour fcst
4:2243950:d=2023092800:ICMR:6 hour fcst
5:2519285:d=2023092800:RWMR:6 hour fcst
6:2800049:d=2023092800:SNMR:6 hour fcst
7:2922175:d=2023092800:GRLE:6 hour fcst
8:2995158:d=2023092800:O3MR:6 hour fcst
...
```

### DIFF
--- a/kerchunk/grib2.py
+++ b/kerchunk/grib2.py
@@ -274,11 +274,12 @@ def scan_grib(
                     )
                     continue
 
-                if idx:
+                if idx and coord in ["time", "valid_time"]:
                     line = ""
                     if coord == "time":
                         time = dt.datetime.fromtimestamp(x, tz=pytz.UTC)
                         line = f"{count}:{offset}:d={time.strftime('%Y%m%d%H')}:{varName.upper()}"
+                        count += 1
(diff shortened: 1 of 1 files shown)


martindurant (commented):
I have no idea of course if the code is right (and there are no tests yet); but what is the use-case for this? If we scan the file, we already know the full information as encoded in the kerchunk references, so why would we want to make an .idx too?

Anu-Ra-g:
I thought local definitions of `.idx` for the grib files might be a good idea. This was just for demo purpose.

martindurant:
> This was just for demo purpose.

That's fine, but please update the PR description or title to say so.

PR #516, opened 2024-10-10
`zarr-python` v3 compatibility

mpiannucci:
So far the only tested file type is HDF5. Thats the only module that currently works with new zarr python in some way


kerchunk/tests/test_hdf.py
```diff
@@ -1,31 +1,50 @@
...
 
 here = osp.dirname(__file__)
 
 
+async def list_dir(store, path):
+    [x async for x in store.list_dir(path)]
+
+
+def create_store(test_dict: dict):
+    if Version(zarr.__version__) < Version("3.0.0.a0"):
```
martindurant, line 28: May as well make the filesystem up front to ensure it's the same for both arms (the else was missing remote_ arguments):
```
fs = fsspec.filesystem("reference", fo=test_dict, remote_protocol="s3", remote_options=so)
```
(and then pass to `RemoteStore` or call `.get_mapper()`)

kerchunk/tests/test_hdf.py
```diff
@@ -1,31 +1,50 @@
...
+import fsspec.implementations.reference
+
 import kerchunk.hdf
 import numpy as np
 import pytest
 import xarray as xr
 import zarr
-import h5py
+
+from packaging.version import Version
```
martindurant, line 15: We could have a global `is_zarr3` somewhere to avoid having to make the comparison in multiple places.

kerchunk/tests/test_hdf.py
```diff
@@ -1,31 +1,50 @@
...
 
 from kerchunk.hdf import SingleHdf5ToZarr, has_visititems_links
 from kerchunk.combine import MultiZarrToZarr, drop
 
 here = osp.dirname(__file__)
 
 
+async def list_dir(store, path):
```
martindurant, line 23: What's this for? I don't think it's used.

kerchunk/hdf.py
```diff
@@ -681,3 +704,44 @@ def _is_netcdf_variable(dataset: h5py.Dataset):
 
 def has_visititems_links():
     return hasattr(h5py.Group, "visititems_links")
+
+
+def encode_fill_value(v: Any, dtype: np.dtype, object_codec: Any = None) -> Any:
```
martindurant: Thanks for writing out all the cases

kerchunk/hdf.py
```diff
@@ -496,6 +516,8 @@ def _translator(
                         if h5obj.fletcher32:
                             logging.info("Discarding fletcher32 checksum")
                             v["size"] -= 4
+                        key =  str.removeprefix(h5obj.name, "/") + "/" + ".".join(map(str, k))
```
martindurant, line 512: This is the same as what `_chunk_key` did? Maybe make it a function with a comment saying it's a copy/reimplementation.

By the way, is `h5obj.name` not actually a string, so you could have done `h5obj.name.removeprefix()`?

kerchunk/hdf.py
```diff
@@ -465,26 +480,31 @@ def _translator(
...
+
+                # Create a Zarr array equivalent to this HDF5 dataset..
+                za = self._zroot.require_array(
+                    name=h5obj.name,
                     shape=h5obj.shape,
                     dtype=dt or h5obj.dtype,
                     chunks=h5obj.chunks or False,
                     fill_value=fill,
-                    compression=None,
+                    compressor=None,
```
martindurant, line 491: So here, you *could* reintroduce the compressor
```
filters = filters[:-1]
compressor = filters[-1]
```
but obviously it depends on whether there are indeed any filters at all.

It would still need back compat, since filters-only datasts definitely exist.

kerchunk/hdf.py
```diff
@@ -157,7 +162,17 @@ def translate(self, preserve_linked_dsets=False):
...
-            store = _encode_for_JSON(self.store)
+            keys_to_remove = []
+            new_keys = {}
+            for k, v in self.store_dict.items():
+                if isinstance(v, zarr.core.buffer.cpu.Buffer):
+                    key = str.removeprefix(k, "/")
+                    new_keys[key] = v.to_bytes()
+                    keys_to_remove.append(k)
+            for k in keys_to_remove:
+                del self.store_dict[k]
```
martindurant, line 167: This is the hacky bit and could use some explanations. Even when requesting "v2", zarr makes Buffer objects, and the keys are also wrong?

kerchunk/hdf.py
```diff
@@ -111,9 +112,14 @@ def __init__(
         if vlen_encode not in ["embed", "null", "leave", "encode"]:
             raise NotImplementedError
         self.vlen = vlen_encode
-        self.store = out or {}
-        self._zroot = zarr.group(store=self.store, overwrite=True)
-
+        self.store_dict = out or {}
+        if Version(zarr.__version__) < Version("3.0.0.a0"):
```
martindurant, line 110: Reuse is_zarr3 mentioned below

martindurant (commented):
There's much less change here than I might have thought.

kerchunk/tests/test_hdf.py
```diff
@@ -23,25 +24,29 @@ async def list_dir(store, path):
...
-def create_store(test_dict: dict):
+def create_store(test_dict: dict, remote_options: Any = None):
     if Version(zarr.__version__) < Version("3.0.0.a0"):
         return fsspec.get_mapper(
-            "reference://", fo=test_dict, remote_protocol="s3", remote_options=so
+            "reference://", fo=test_dict, remote_protocol="s3", remote_options=remote_options
         )
     else:
-        fs = fsspec.implementations.reference.ReferenceFileSystem(fo=test_dict)
+        fs = fsspec.implementations.reference.ReferenceFileSystem(fo=test_dict, remote_options=remote_options)
```
martindurant, line 33: + remote_protocol

kerchunk/hdf.py
mpiannucci: yeah the big issue is that v3 cares about what type of operation it is, and v2w doesnt so moving them around doesnt necessarily fix that bug

kerchunk/hdf.py
mpiannucci: Yeah so two issues here: 

1. the keys we get from hdf are for example `/depth/.zarray` when then need to be `depth/.zarray`
2. we cant jsonify buffers, which is how the internal `MemoryStore` in v3 stores its data. So we need to convert the buffers to bytes to be serialized

martindurant: OK - would appreciate comments on the code saying this.

kerchunk/hdf.py
martindurant: So there needs to be a change upstream?

mpiannucci: Yes this: https://github.com/zarr-developers/zarr-python/issues/2325

mpiannucci:
Also worth noting... zarr 3 doesn't support numcodecs codecs out of the box. There is a pr to help this https://github.com/zarr-developers/numcodecs/pull/524 () but it would mean a change to codec names which causes an incompatibility. For the initial icechunk examples we handle this in virtualizarr but long term it probably belongs here to work standalone.

mpiannucci:
When this https://github.com/zarr-developers/zarr-python/pull/2425 goes in it should unblock this to work full with zarr python v3. 

We will also need to create both numcodecs and zarr v3 codec versions of all the custom kerchunk codecs so that a given dataset can be loaded in either v2 or v3 contexts (say if yuou kerchunk a grib file, then want to convert those references to an icechunk store

martindurant:
> create both numcodecs and zarr v3 codec versions of all the custom kerchunk codecs

Is that a subclass in each case, specifying that it is to be bytes->bytes?

mpiannucci:
Sorry the numcodecs versions exists as they are today. Yes the v3 version would be basically using the numcodecs versions. 

Although grib decoder should really be bytes to array i think

mpiannucci:
This is required upstream: https://github.com/fsspec/filesystem_spec/pull/1734

mpiannucci:
Also of note: To get this to work with zarr 3, we pass an fsspec ReferenceFilesystem to a zarr RemoteStore. This works fine with remote filesystems (where the references live) that have async implementations (s3, http, etc) but does not work when the references are on filesystem with only a sync implementation (local, etc). The tests heavily depend on the local filesysmte, and im sure many others do as well so it needs to be figured out how the interaction of sync filesystems and the async zarr RemoteStore requirement work out

mpiannucci:
Grib works as long as read in zarr 2 format. 

To be used with zarr 3 the codec needs to be ported over to the new [abc.Codec](https://github.com/zarr-developers/zarr-python/blob/main/src/zarr/abc/codec.py) class from zarr

mpiannucci:
I think one change that should be made for compatibility with zarr 3 and virtualizarr is that the Grib2 codec moves from filters to compressors.

martindurant:
>  the Grib2 codec moves from filters to compressors

It returns an array, though - this was already pointed out. I think there is an assumption that compressors map to byte->byte codecs.

mpiannucci:
Ill change it back for now. Will discuss with zarr folks

mpiannucci:
State of the Union post merge with @moradology and installing latest fsspec and zarr 3

```
pip install git+https://github.com/fsspec/filesystem_spec
```

- `test_grib.py`: 1 Failure
- `test_netcdf.py`: 3 Failures
- `test_hdf.py`: 11 Failures
- `test_df.py`: ✅ 
- `test_utils.py`: 3 Failures
- `test_fits.py`: 5 Failures
- `test_zarr.py`: 2 Failures
- `test_combine.py`: All Failed
- `test_combine_concat.py`: 8 Failures
- `test_combine_dask.py`: 3 Failures
- `test_xarray_backend.py`: 1 Failure

A lot of these are the same errors over and over a again. The hardest part will be maintaining compat for zarr python 2 library which I am not sure if it should be a goal of this PR @martindurant

maxrjones:
Thanks for working on this compatibility! I really appreciate all the efforts you've put into this already. My understanding is that there's still a bit of work to do. Even though we all know that Matt's super trustworthy, some concerns were raised in the VirtualiZarr coordination meeting yesterday about how the dependency conflicts between icechunk, virtualizarr, zarr, and kerchunk seem more urgent with Zarr v3.0 out and it being sketchy to recommend  installing a library from a branch off a fork. I wanted to float the idea of either moving Matt's PR to a branch within fsspec/kerchunk or adding a fsspec/kerchunk@zarr-v3 branch to mitigate these concerns until zarr-python v3 compatibility is completed. what do y'all think?

moradology:
Makes sense to me. Perhaps a similar target branch should be added across the impacted repos (assuming that doesn't increase the cognitive load too much)

martindurant:
It is still the case that kerchunk can be used to generate v2 output with zarr-3 installed, no?
Currently, the pyproject requirement has zarr<3, so in practice, kerchunkers should not be using the newly-released one, but the output ought to still be readable by it.

Did I misunderstand something? Please invite me to the coordination meeting, so we can decide who does what and when.

martindurant:
I see I am wrong, and this branch *cannot* be expected to work with v2 zarr. So the plan is for all future kerchunking to happen in v3 (but perhaps still produce v2 output, so that people who don't upgrade can still use it) ?

mpiannucci:
That is correct

maxrjones:
>  Please invite me to the coordination meeting, so we can decide who does what and when.

@martindurant the coordination meeting info is at https://zarr.dev/community-calls/. The event is hosted on Zarr's community calendar rather than a regular calendar invite so it's not tied to any single individual/institution.

martindurant:
OK, so I can work on this, if you have no time @mpiannucci . I will go with the plan, that kerchunk will require zarr>3, but produce v2 output (because we don't actually use any new functionality), and for combining and zarr-to-zarr, the input has to be v2 format. 
zarr>3 also puts a condition on the minimum python version, I think >=3.11.

martindurant:
Status:
```
tests/test__grib_idx.py FFFF.FF                                                                                                                                                         [  6%]
tests/test_combine.py ....FFFFF....FFF.......F...FFFF.....                                                                                                                              [ 37%]
tests/test_combine_concat.py ........                                                                                                                                                   [ 43%]
tests/test_combine_dask.py ...                                                                                                                                                          [ 46%]
tests/test_df.py ..                                                                                                                                                                     [ 48%]
tests/test_fits.py ......                                                                                                                                                               [ 53%]
tests/test_grib.py .s.............sss                                                                                                                                                   [ 68%]
tests/test_hdf.py ..............FFF                                                                                                                                                                                       [ 83%]
tests/test_netcdf.py ...                                                                                                                                                                [ 86%]
tests/test_tiff.py ..                                                                                                                                                                   [ 87%]
tests/test_utils.py ..........                                                                                                                                                          [ 96%]
tests/test_xarray_backend.py .                                                                                                                                                          [ 97%]
tests/test_zarr.py ...                                                                                                                                                                  [100%]
```

The HDF ones are critical deadlocks that will be hard to fix. The remaining combine ones are probably not hard. The grib ones are specific to xarray-datatree to do with filepath munging.

martindurant:
OK, only failures now are in test__grib_idx.py , which use datatree. I *think* this is to do with a leading "/" in the paths...

maxrjones:
> OK, only failures now are in test__grib_idx.py , which use datatree. I _think_ this is to do with a leading "/" in the paths...

🎉 exciting progress! DataTree support is not yet supported fully for Zarr-Python 3, see https://github.com/pydata/xarray/issues/9984. I think a conditional xfail would be the way to go here.

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   "numcodecs",
   "numpy",
   "ujson",
-  "zarr<3",
+  "zarr",
```
maxrjones, line 27:
```suggestion
  "zarr>=3",
```
I think you'll  need to pin this to only >3 because a lot of the new syntax won't work with the 2.x API

martindurant: Yes, agreed.

martindurant:
I don't know how much effort we want to spend finding the deadlock triggered by the HDF5 embed test that is now commented out.

moradology:
I'd been bashing my head against those grib tests all day and couldn't figure out why, even in a debugger and doing things manually, none of the DataTree nodes had any data...

In the process of messing with this I uncovered what may be a `FsspecStore` bug in its `list` method: if the path it is given is "", it will replace all "/" with "" and return some very odd results e.g. 
```Store keys: ['.zattrs', '.zgroup', 'u.zattrs', 'u.zgroup', 'uinstant.zattrs', 'uinstant.zgroup', 'uinstantheightAboveGround.zattrs', 'uinstantheightAboveGround.zgroup', 'uinstantheightAboveGroundheightAboveGround.zarray', 'uinstantheightAboveGroundheightAboveGround.zattrs', 'uinstantheightAboveGroundheightAboveGround0', 'uinstantheightAboveGroundlatitude.zarray', 'uinstantheightAboveGroundlatitude.zattrs', 'uinstantheightAboveGroundlatitude0.0', 'uinstantheightAboveGroundlongitude.zarray', 'uinstantheightAboveGroundlongitude.zattrs', 'uinstantheightAboveGroundlongitude0.0', 'uinstantheightAboveGroundstep.zarray', 'uinstantheightAboveGroundstep.zattrs', 'uinstantheightAboveGroundstep0', 'uinstantheightAboveGroundtime.zarray', 'uinstantheightAboveGroundtime.zattrs', 'uinstantheightAboveGroundtime0', 'uinstantheightAboveGroundu.zarray', 'uinstantheightAboveGroundu.zattrs', 'uinstantheightAboveGroundu0.0.0.0', 'uinstantheightAboveGroundvalid_time.zarray', 'uinstantheightAboveGroundvalid_time.zattrs', 'uinstantheightAboveGroundvalid_time0.0']```

moradology:
Might want to add this small update as well, since it better captures the intent of the version check so far as I can tell: https://github.com/mpiannucci/kerchunk/pull/6

martindurant:
> In the process of messing with this I uncovered what may be a FsspecStore bug in its list method: if the path it is given is "", it will replace all "/" with "" and return some very odd results

!!! Please file this as an issue. Maybe it helps with the datatree thing too.

maxrjones:
> I'd been bashing my head against those grib tests all day and couldn't figure out why, even in a debugger and doing things manually, none of the DataTree nodes had any data...
> 
> In the process of messing with this I uncovered what may be a `FsspecStore` bug in its `list` method: if the path it is given is "", it will replace all "/" with "" and return some very odd results e.g. `Store keys: ['.zattrs', '.zgroup', 'u.zattrs', 'u.zgroup', 'uinstant.zattrs', 'uinstant.zgroup', 'uinstantheightAboveGround.zattrs', 'uinstantheightAboveGround.zgroup', 'uinstantheightAboveGroundheightAboveGround.zarray', 'uinstantheightAboveGroundheightAboveGround.zattrs', 'uinstantheightAboveGroundheightAboveGround0', 'uinstantheightAboveGroundlatitude.zarray', 'uinstantheightAboveGroundlatitude.zattrs', 'uinstantheightAboveGroundlatitude0.0', 'uinstantheightAboveGroundlongitude.zarray', 'uinstantheightAboveGroundlongitude.zattrs', 'uinstantheightAboveGroundlongitude0.0', 'uinstantheightAboveGroundstep.zarray', 'uinstantheightAboveGroundstep.zattrs', 'uinstantheightAboveGroundstep0', 'uinstantheightAboveGroundtime.zarray', 'uinstantheightAboveGroundtime.zattrs', 'uinstantheightAboveGroundtime0', 'uinstantheightAboveGroundu.zarray', 'uinstantheightAboveGroundu.zattrs', 'uinstantheightAboveGroundu0.0.0.0', 'uinstantheightAboveGroundvalid_time.zarray', 'uinstantheightAboveGroundvalid_time.zattrs', 'uinstantheightAboveGroundvalid_time0.0']`

I can open a PR to fix this in Zarr-Python, but agree that an issue will be good. thank you for finding this bug!

TomNicholas:
Great to see this progress!

The preceding `/` bug was discussed at length in the Xarray meeting earlier today. We will fix it, but if you can xfail it for now then that would suit us in VirtualiZarr.

martindurant:
@TomNicholas , I am skipping that whole test module - all is green

TomNicholas:
Wow! Okay please let us know when you release this then ❤️

moradology:
@martindurant Not a huge deal, but I noticed that this PR didn't make it in: https://github.com/mpiannucci/kerchunk/pull/6

Should I cut it against `kerchunk` main or nah?

martindurant:
Actually, it looks like that's now dead code. Following this PR, kechunk *only* works with zarr3, for better or worse.

PR #523, opened 2024-10-28
remaining corpus of idx mapping functionality

emfdavid:
This is the remainder of the idx mapping code - primarily reinflate_grib_store and supporting methods that allow creating datasets from the idx files.

This PR is pretty large. I think the notebook example should go to https://github.com/ProjectPythia/kerchunk-cookbook ?
But I wanted you to be able to see the rest and how it works. Happy to break it up however you want. I know the methods need better doc strings, but I wanted you to give me a little more feedback on how to proceed first.

There are also pretty extensive [tests](https://github.com/asascience-open/nextgen-dmac/blob/main/grib_index_aggregation/test_dynamic_zarr_store.py) and I will move those too but they require a bit more work to convert from unittest to pytest.


rsignell:
@martindurant , just to make sure I understand, this is awaiting your review?

martindurant:
Yes, but it's only 12hr old :)  @emfdavid is asking for feedback above, so if you are in a place to do that, we would appreciate that too.

rsignell:
@martindurant, sorry, didn't mean to suggest impatience!   🙃 
I will try giving it a test right now!

emfdavid:
@rsignell I am impatient 😆  did you kick the tires yet?

rsignell:
@emfdavid , heh heh, sorry, got distracted with another burning fire, but I just tried it out.  Running into some xarray/datatree issue, perhaps because the latest release (2024.10.0) of xarray has datatree built in now and there is some conflict if you try to not use the built in version?
```
ImportError: cannot import name 'HybridMappingProxy' from 'xarray.core.utils' (/home/rsignell/miniforge3/envs/grib-idx/lib/python3.13/site-packages/xarray/core/utils.py)
```

martindurant:
The CI is showing that too, so yes, probably need to pin xarray or not-install tree

rsignell:
Ideally this code would use the now-built-in xarray datatree -- would that be an easy fix for you @emfdavid ?

rsignell:
So I did manage to run the test notebook using `xarray=2024.7.0` but had to change `ulwrf/avg/nominalTop` to `sulwrf/avg/nominalTop`.  Do the data at the end look okay?   Seem a bit strange...
https://gist.github.com/rsignell/f1eb9cfc4d8b9cc88c3ac9bda2504f9a

emfdavid:
I updated the code to use the now built in datatree 🎉 
Excited that is merged into xarray now!

I can't run your notebook tonight but those nan values do look funny. Building the axes is really finicky... I can try to take a look tomorrow.

I did run down the issue with the variable names dswrf -> sdswrf, ulwrf -> sulwrf
It looks like eccodes [updated](https://github.com/ecmwf/eccodes/commit/9f0432deb0d0f57c8ac002b2d14236980d444674#diff-bf52f84770b011c828396c842f5a282899bede6bb3acd2c967b50eabca327e42L164-R165) their definitions.

Is that something we can ask them to undo?
I will be inclined to pin cfgrib/eccodes till that is confirmed... but the mapping based approach works just fine either way!

rsignell:
I was able to run a [slightly-modified version of the notebook](https://nbviewer.org/gist/rsignell/bf3ca5cde078153f5d69e5cd86966d7b) using the latest xarray (2024.10.0) and https://github.com/fsspec/kerchunk/pull/523 !

When I extracted some data values from the virtual dataset, however, they seem unlikely to be correct.

Could you please check the data values in the last cell?   
Perhaps we should hold off merging until we decide whether the data extraction is working properly...

emfdavid:
I will see if I can debug your notebook tonight. If you have a chance to try the one in this PR I would appreciate it.

As far as when to merge... do you (@martindurant) want the tests added to this already large PR?
https://github.com/asascience-open/nextgen-dmac/blob/main/grib_index_aggregation/test_dynamic_zarr_store.py
That will take a bit more work to convert from unittest to pytest but should be straight forward enough?

emfdavid:
Here is the state of the art on huggingface... I would be really excited to see an update here!
Screenshot 2024-10-29 at 9 16 57 AM

rsignell:
@emfdavid okay, I'm testing the notebook in the repo now...

Anu-Ra-g:
@rsignell In this [notebook](https://gist.github.com/Anu-Ra-g/e0d7b9c815dcd4df4ea2261ed272e5d3), I was able to reinflate the generated indexes but I missed to look at the actual data in those indexes.

Anu-Ra-g:
> I was able to run a [slightly-modified version of the notebook](https://nbviewer.org/gist/rsignell/bf3ca5cde078153f5d69e5cd86966d7b) using the latest xarray (2024.10.0) and your latest fixes to #523 !
> 
> When I extracted some data values from the virtual dataset, however, they seem unlikely to be correct.
> 
> @emfdavid, @Anu-Ra-g, or @martindurant, is there a simple alternate method by which we could check to see if the data values in the last cell are correct?
> 
> Perhaps we should hold off merging until we decide whether the data extraction is working properly?

The data extraction is working properly as there are minimal code changes and refactors from the original code. But I used **GEFS** grib data for making that notebook. Maybe that could be a reason! 

As per the original demonstrations by @emfdavid with the **GFS** data, the data extraction works fine. 

If you want, I could change it from **GEFS** to some other model's data.

emfdavid:
@rsignell There is a white space error in the double for loop. The notebook is only processing the 18z forecast for each day which is why there are three nan value between each real value.

Screenshot 2024-10-29 at 10 32 26 PM

I also noticed there are a lot of duplicate attrs in that much older 2017 file. The GEFS model changed significantly around 2020-09-25. I think these are really issues with the NOAA Grib file and its compatibility with cfgrib.
Screenshot 2024-10-29 at 10 07 54 PM
But these UGRD variables all reading out on the same grib level=0.0 are going to be a problem.

I don't see this on the more recent GEFS data, but I have only really looked at the geavg files so far.

emfdavid:
@martindurant I tried to fix the build issue, with ImportError "HybridMappingProxy" for just the 3.10 build but I think I made it worse with [d40cca4](https://github.com/fsspec/kerchunk/pull/523/commits/d40cca4e18d98d2a3f7e219f439176438c81bae8)
Any suggestions on what to do here? I think maybe conda doesn't have a xarray 2024.10 for python 3.10?

We can definitely rebase away the notebooks before merging - those are not to be checked in here like this.

rsignell:
Whoa, great find on that white space indent error @emfdavid !    The notebook indeed works fine now: https://nbviewer.org/gist/rsignell/c3fd58368ed9d0ae50c26807b6a51678

martindurant:
OK, I am happy to push this in, as is - it's fine to include the example notebook, and we can always iterate yet. OK, everyone?

emfdavid:
Sure - your call on the notebook - it is a bit chunky to have in the git tree at ~12mb...
Let me know and I can remove it or you can merge as is.

I will work on the tests shortly. There will be some moderate size test fixture files for that.
Does [this](https://github.com/asascience-open/nextgen-dmac/tree/main/grib_index_aggregation/fixtures) structure look okay?

martindurant:
Can we put in the un-evaluated notebook?

emfdavid:
Pushed the change - can you squash merge?
If not, I can rebase the old commits away.

PR #525, opened 2024-10-30
Fix coo_map when only one match from regex pattern

aaTman:
I ran into an issue when using `coo_map` in `MultiZarrToZarr` that was caused by using `.groups()[0]` instead of `.group()`, where there is only one returned regex match in the string (at least, as far as I can tell is the case).

To fix this, I kept the original functionality but would catch if the `.groups()` tuple was empty. In the situation I ran into this bug, I used the following code which caused an IndexError due to a length 0 tuple:

```
pattern = re.compile(r"[A-Za-z]\d\d(?![^ ]*[\\\/])", re.IGNORECASE)
file_list = glob.glob(f"{self.directory}/*")
mzz = MultiZarrToZarr(
    file_list,
    coo_map={"member": pattern},
    concat_dims=["member", "step", "time"],
    identical_dims=["latitude", "longitude"],
)
multi_kerchunk = mzz.translate()
```

The string (and similar ones) that returned only one matching pattern to the regex was most recently:

`/var/folders/vz/txd62qzn76g9f6cxg_8_76cw0000gn/T/tmp_kwkysl7/pres_msl_2002110200_p03_01.json`

Where my goal was to subset the "p03" and other ensemble members from the GEFSv12 Retrospective data. 

It seems this completely fixed the issue though happy to discuss or try some other edge cases.


martindurant:
These changes I think should fix the datatree thing, as well as requiring the current version of xarray: https://github.com/fsspec/kerchunk/pull/523/files#diff-65c089582c22ac69bff6e677f8c5dcf10b650de2fac7193ed15b013c7866925d

PR #537, opened 2025-01-20
sync test?

emfdavid:
Whats the issue?

### DIFF
--- a/kerchunk/utils.py
+++ b/kerchunk/utils.py
@@ -8,6 +8,7 @@ import ujson
 import fsspec
 import zarr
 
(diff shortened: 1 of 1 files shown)


martindurant:
It seems like something that one might see from zarr 3, but here we use v2. I think we'll be updating everything to work with zarr 3 ( https://github.com/fsspec/kerchunk/pull/516 ), should be ready relatively soon.

emfdavid:
@martindurant can we move forward with #528 since the issue is unrelated?

martindurant:
If you say it's ready, it's ready.

PR #541, opened 2025-01-30
Add more specific version requirements for Zarr in pyproject.toml

maxrjones:
Zarr Python 3 is needed, not >3

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   "numcodecs",
   "numpy",
   "ujson",
-  "zarr>=3",
+  "zarr>=3.0.1",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,7 @@ dependencies = [
   "numcodecs",
   "numpy",
   "ujson",
-  "zarr>3",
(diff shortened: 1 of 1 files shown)


martindurant:
I'm not sure we wanted to allow ==3.0.0, I think we are depending on some fairly cutting edge things here

TomNicholas:
Isn't the problem here that >3 implies zarr 4?? IIUC should either be >=3 or >3.0.0

maxrjones:
> Isn't the problem here that >3 implies zarr 4?? IIUC should either be >=3 or >3.0.0

This is what I assumed too, but I think Martin could be right that 3.0.1 satisfies >3. I'm trying to find a primary source of guidance now.

TomNicholas:
> I think we are depending on some fairly cutting edge things here

Then just pin to the latest released version of zarr-python - the one you're presumably testing against? (Which is apparently only 3.0.1 anyway)

maxrjones:
I updated it to >=3.0.1 since it seems there was not typo, but being explicit about version requirements is helpful if kerchunk does rely on fixes between 3.0.0 and 3.0.1.

TomNicholas:
That also seems consistent with what every other package that depends on zarr-python v3 is doing.

PR #560, opened 2025-06-05
Issue 559: Update scan_grib to store levels with two surfaces

BjarkeMoholt:
Update scan_grib to include evaluation of two-surface levels when determining if a message has levels


martindurant:
I'm fine with this if it fits your use case. Can we include a docstring or comment in the new utility function to explain the why?

Also, if you have a *small* public example file, we could include it for a test.

PR #565, opened 2025-07-11
Ignore errors when decoding attributes in netCDF3.py

mannreis:
Faced some issues when using virtualizarr because of this

### DIFF
--- a/kerchunk/netCDF3.py
+++ b/kerchunk/netCDF3.py
@@ -216,7 +216,7 @@ class NetCDF3ToZarr(netcdf_file):
                 ]
             arr.attrs.update(
                 {
-                    k: v.decode() if isinstance(v, bytes) else str(v)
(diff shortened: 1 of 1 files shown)


martindurant:
If you have a link to an issue on vzarr, would appreciate including it here

martindurant:
Tests appear broken by recent zarr release, which requires type to go Object->str in HDF. Would you have time to find and fix this?

martindurant:
> I managed to reproduce the problem locally but by upgrading zarr to 3.0.10 the issue goes away

!!

So actually he only thing that's wrong here is the failing tests. I'll merge this and try to fix it, although this week is rammed.

mannreis:
In the meantime the `zarr-python/main` version is `3.1.0` so there's a change retriggering the tests will make them pass @martindurant.

I've opened https://github.com/fsspec/kerchunk/pull/567 as side follow up on this testing debugging. Feel free to close it if it's not really desired to tests against a "stable" zarr release

PR #589, opened 2026-09-15
Add maintenance-mode notice pointing to VirtualiZarr

TomNicholas:
After a conversation with @martindurant, I suggested we add this banner to clarify the development status of Kerchunk to all users.

Kerchunk is effectively no longer under active feature development, since (paraphrasing @martindurant) the core developer (i.e. Martin) is no longer able to devote significant time to the project.

As discussed via DM, this PR adds a note to the README and a dismissible banner across the docs site pointing new projects to VirtualiZarr and Icechunk as alternatives. The banner looks like this:

Screenshot 2026-09-15 at 3 24 59 PM


martindurant:
cc @FrancescAlted

TomNicholas:
FYI @martindurant if you simply merge this PR the docs should rebuild off it and the warning will be visible to users, there is no requirement to issue a release to make it visible. (And there are tests failing on main already so that's unrelated.)

TomNicholas:
Thank you @martindurant !

